PR #24, opened 2015-07-28
Prototype of a solution for string with zeros.

cypof:
We are having trouble using the Caffe wrapper when a C++ string contains a zero. Instead of using the string size, JavaCPP uses the C convention and cuts it. This is a very hackish fix that seems to solve this for now. I am interested to figure out a more general fix.

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/tools/Generator.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Generator.java
@@ -516,14 +516,14 @@ public class Generator implements Closeable {
             out.println("class JavaCPP_hidden StringAdapter {");
             out.println("public:");
             out.println("    StringAdapter(const          char* ptr, size_t size) : ptr((char*)ptr), size(size),");
-            out.println("        str2(ptr ? (char*)ptr : \"\"), str(str2) { }");
+            out.println("        str2(ptr ? (char*)ptr : \"\", size == (size_t) (jint) -1 ? strlen((char*)ptr) : size), str(str2) { }");
             out.println("    StringAdapter(const signed   char* ptr, size_t size) : ptr((char*)ptr), size(size),");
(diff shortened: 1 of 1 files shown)


saudet:
Thanks checking this out! I've thought about other areas in the code, and I think I was able to come up with something that works well enough. Let me know what you think, thanks!

cypof:
I will try this right now, thanks

cypof:
I thought this was working but it seems strings still get truncated on master. Would you be interested to work on trying to simplify the adapter? It's pretty hard to figure out what is going on, and having the char* operator change the state of the object is confusing. For example why is there both a string, string& and char*/size fields at the same time? Thanks

saudet:
Could you provide a test case that fails? I'll get back to you on the rest
later. Thanks!

saudet:
`StringAdapter` is a hack to sort of transparently bind `std::string` with it's Java equivalent, `java.lang.String`, which is immutable. So, I've been thinking about this today, one thing it won't do that might be causing problems here is the it won't return a `new std::string` pointer that can survive after `StringAdapter` gets destroyed. It assumes the content is going to get copied. That might not work for data, where we have the need to modify the content of some object that gets provided to us. In that case, it would be best to provide a separate wrapper class, named something like `DataString`.

`StringAdapter` is confusing, yes, and I'm sure we could do better. It's just not something anyone has ever tried to do before, so it's hard to figure out everything on my own. Sometimes the user provides us a `std::string` so we want to use that as container, but sometimes we get a `char*` and we need to create a `std::string` of our own. I'm not sure how to go about this in a better way. I unfortunately don't have any better ideas at the moment. If you have any suggestions, please do let me know. :) Thanks

In any case, please let me know where things are failing, either as part of Caffe or a separate test case, and I'll see what we can do about that, thanks!

PR #26, opened 2015-07-28
Check if object creation failed, happens for abstract classes.

cypof:
Not sure if the exception should be cleared. JavaCPP currently segfaults a bit later, I guess it would be better to raise the exception on the Java side.


saudet:
`AllocObject()` returns `NULL` when it fails, and it usually gets called just before returning to Java (but not always, might have to study this a bit further), so we might as well let it propagate the exception back to Java. I fixed a couple of missing `NULL` checks in the commit above, and now when calling `FloatLayer.layer_by_name()`, for example, we get a proper stack trace like this:
```
java.lang.InstantiationException: org.bytedeco.javacpp.caffe$FloatLayer
	at org.bytedeco.javacpp.caffe$FloatNet.layer_by_name(Native Method)
	at caffe$3.command(caffe.java:238)
	at caffe.main(caffe.java:392)
```
Is this satisfactory?

cypof:
Yes that's great, a java exception is better than just a message, thanks

PR #40, opened 2015-09-19
Ignore OSX and IDE files

cypof:


saudet:
BTW, could you do the same for the `javacv` repo as well? Thanks!

Issue #61, opened 2016-01-11
Duplicate code generated


243826:
I am not able to pinpoint what I am doing wrong. I have a class (CPPOutputPort) which is referenced from 2 different java classes. I see that the CPPOutputPort class has repeated definitions for peer cpp code in the generated cpp file (target/jni/CandidApex.cpp). 

You need mvn installed in order to build the project using the following  command. Alternately the pipeline/pom.xml shows how it's invoking javacpp.

```
unzip pipeline.zip && cd pipeline && mvn package
```

[pipeline.zip](https://github.com/bytedeco/javacpp/files/86515/pipeline.zip)

243826:
On the flip side, if I spell out the 2 classes underneath the routing package instead of package glob, then the definition for Javacpp_CPPOutputPort is completely omitted.

saudet:
The archive is missing a few header files, so it's hard to test out. What are the repeated definitions that you get? I see a couple of them around `JavaCPP_CPPOutputPort`, is that all or is there more?

243826:
The missing header file is needed to compile the generated code. You can generate the code without it though. The problem is with the code generation. The generated code has repeated declaration and definition for JavaCPP_CPPOutputPort. That's all. Please look at my pull request.

243826:
when do you close the issue? After release or after merge?

saudet:
I tend to close issues that have remained opened after a release, as a sort of review, but people who are already satisfied anyway close issues by themselves right away after it's in master, your choice. :)

saudet:
Released in version 1.2! Thanks for the bug report

PR #62, opened 2016-01-12
#61 dedup all the classes for which the code needs to be generated

243826:


saudet:
Is this something that happens with the sample `pipeline.zip` you posted in issue #61?

243826:
yes. when I provide package name as argument, it detects that CPPOutputPort.class needs to be included as well _twice_. I assume once per class in the package.

saudet:
I see, your changes seem to fix the issue alright. So, we should be using a `Set` there instead of a `List` from the start. What about changing the type of `map` to `LinkedHashMap<String, Set<Class>>` and putting `new LinkedHashSet<Class>()` into that instead of `new ArrayList<Class>()`? Plus, that should preserve the order in which the classes are currently returned.

243826:
I had the same thought about LinkedHashSet. It's a quick fix - I will augment the PR.

Regarding "using Set from the start", I actually started down that path but soon realized that my changes were actually percolating into a huge change set in the code which I hardly understood. I am not comfortable doing it myself.

saudet:
Actually, no, it's not a big change. The map object is only used in that
method. It doesn't go outside!

243826:
You are right. I have updated the PR.

saudet:
Looks great! Thanks a lot :)

PR #67, opened 2016-01-29
squid:S00108 - Nested blocks of code should not be left empty

georgekankava:
This pull request is focused on resolving occurrences of Sonar rule squid:S00108 - Nested blocks of code should not be left empty.
You can find more information about the issue here:
https://dev.eclipse.org/sonar/coding_rules#q=squid%3AS00108
Please let me know if you have any questions.
George Kankava

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/Loader.java
+++ b/src/main/java/org/bytedeco/javacpp/Loader.java
@@ -150,12 +150,16 @@ public class Loader {
             } finally {
                 try {
                     is2.close();
-                } catch (IOException ex) { }
+                } catch (IOException ex) {
+                    logger.error("Unable to close resource : " + ex.getMessage());
+                }
             }
         } finally {
             try {
                 is.close();
-            } catch (IOException ex) { }
+            } catch (IOException ex) {
+                logger.error("Unable to close resource : " + ex.getMessage());
+            }
         }
         return p;
     }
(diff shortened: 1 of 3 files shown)


saudet:
Thanks for the recommendations! A couple of things though:
1. I've dealt with `InterruptedException` in a different way for the issue at https://github.com/bytedeco/javacv/issues/315.
2. The comment in `Parser` for `IllegalAccessException` actually applies to the two other types of exceptions as well, so I think it's alright the way it is now?
3. On error, `Pointer.offsetof()` returns -1, so we don't need to log anything. I guess we could put `return -1` in the catch block to make this more explicit?
The rest is fine, so I'll be happy to merge that once adjusted! Thanks

georgekankava:
Fixed. Thanks!

saudet:
Could you make one commit out of this? The easiest way is to `git reset master` your branch, recommit, and `git push --force`.

And do the same for the other other pull requests! Thanks

georgekankava:
Done Sir.

saudet:
Thanks! But there is still a problem with the "fail silently" part. It isn't supposed to report or log anything. Would you have a suggestion about how to make that fact clearer for all the types of exceptions thrown there, without repeating the comment three times, if possible? thanks!

saudet:
Any updates?

georgekankava:
Why not to catch general Exception not to repeat empty blocks 3 times or Java 7 multi catch clause can also be used.

saudet:
Multicatch is a good idea I think! It should have no problem on Android as
well.

georgekankava:
Please take a loot at changes. Hope this is what was expected.

saudet:
Looks good, thanks!

PR #74, opened 2016-02-16
squid:S2259 - Null pointers should not be dereferenced

georgekankava:
This pull request is focused on resolving occurrences of Sonar rule
squid:S2259 - Null pointers should not be dereferenced.
You can find more information about the issue here:
https://dev.eclipse.org/sonar/coding_rules#q=squid%3AS2259
Please let me know if you have any questions.
George Kankava


saudet:
Lines 598 and 1248 cannot lead to dereferencing null, and your fix on line 600 doesn't compile. For the former, could you improve on making that fact clear to static code analyzers without having to check things twice? Thanks!

saudet:
To be clear, line 600 is problematic, and we should add a check...

PR #77, opened 2016-02-21
add definition of linux-armhf arch based on sun.arch.abi property

vb216:
Trying to add a clean addition that won't break anyone already using linux-arm for other arm types, but will allow specification of flags aimed at the pi2, so platform specific linux-armhf jars could be cross-compiled for that target

Update for cppbuild.sh in javacpp-presets would build these dependencies, and then the Loader class select to use them on the target


saudet:
LGTM, thanks!

PR #82, opened 2016-03-08
long pointers

agibsonccc:
Adds long pointers - WIP.  Need to resolve jni adapter issues yet.

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/BoolPointer.java
+++ b/src/main/java/org/bytedeco/javacpp/BoolPointer.java
@@ -40,7 +40,7 @@ public class BoolPointer extends Pointer {
      *
      * @param size the number of {@code bool} elements to allocate
      */
-    public BoolPointer(int size) {
+    public BoolPointer(long size) {
         try {
             allocateArray(size);
         } catch (UnsatisfiedLinkError e) {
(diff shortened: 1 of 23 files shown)


agibsonccc:
Errors associated with it: https://gist.github.com/agibsonccc/e6d647717da28f344cf6

saudet:
I was worried that because 64-bit arithmetic in Java doesn't get optimized very well that maybe `GetLongField()` would be slower than `GetIntField()` even on 64-bit architecture, but I've confirmed that this isn't the case. They both run in ~5 ns with OpenJDK 8. So, it doesn't look like we would suffer any performance degradation. The memory footprint of each Pointer almost doubles though, but it looks like that would be the only downside. And even there, it would only be significant when keeping a large amount of Pointer objects in memory that refer to small areas of memory, and I don't believe anyone currently needs to use that way.

agibsonccc:
What if you have a "mode" for pointer memory? Or maybe have a new module/pointer factory? You could implement both over loading of ints and longs.

saudet:
I don't think it's worth the trouble, that's the point. I have to give this a little bit more thought though, obviously. :)

saudet:
Merged into the longptr branch: https://github.com/bytedeco/javacpp/tree/longptr

PR #84, opened 2016-03-11
Mark functions from c includes with noexception

osialr:
This allows the Generator to avoid wrapping the function invocations with
try/catch when no exception is possible


saudet:
GCC doesn't do these kinds of optimization by default because we might have callbacks that throw C++ exceptions, but I think it's a good idea to offer the option to the user, and this looks like a good default.

Could you only remove the '\n' and replace that with a white space like it's done for other annotations? And I'll merge that in. Thanks!

osialr:
Amended with your suggestions

PR #85, opened 2016-03-11
Add flag to keep the generated cpp files

osialr:
I would like to inspect and deliver the generated cpp functions along with the libraries.


saudet:
Good idea :) But I would see this as a `-nodelete` option for the builder. What do you think? Would it be possible to place this new feature alongside the `-nocompile` option in `Builder` as well as the corresponding `compile` property in `BuildMojo`?

osialr:
Your way makes much more sense.  I'll work on redoing it.

PR #86, opened 2016-03-11
Remove 'throws Exception' from Pointer::close

osialr:
Pointer throws a RuntimeException on close.  Removing the throws
allows the user's try-with-resource blocks not require a catch(Exception)

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/Pointer.java
+++ b/src/main/java/org/bytedeco/javacpp/Pointer.java
@@ -402,7 +402,7 @@ public class Pointer implements AutoCloseable {
     }
 
     /** Calls {@code deallocate()}. */
-    @Override public void close() throws Exception {
+    @Override public void close() {
         deallocate();
     }
 


saudet:
I've thought about that myself, but if we do that, then it becomes impossible to override the `close()` method with a checked exception, like `IOException`, so everything would need to be wrapped in a `RuntimeException` or something. Is this really wise?

On the other hand, it *is* possible to override the `close()` method to get rid of the `throws Exception`, so depending on the need of each class, it seemed to me that this way the user was less limited, for example:

```java
public NoThrowClass extends Pointer {
    @Override public void close() { super.close(); }
}
```
```java
public CustomThrowClass extends Pointer {
    @Override public void close() throws CustomException { super.close(); }
}
```

osialr:
I see your point and that adding this could be a breaking change.  My reasoning was that `deallocate()` can't throw `Exception` so I propagated that same attribute up to `close()`

A custom class could override `close()` to remove the `throws`.  But the auto-generated are still stuck with `Exception`.

The `Parser` could add a `throw`-less close to each autogenerated struct:
 ```java
@Override
void close() { 
    try {
        super.close();
    } catch (Exception e) { 
       throw new RuntimeException(e);
    }
}
```

**Comment by saudet** [comment]
Right, `super.close()` throws, so actually:
```
@Override public void close() { deallocate(); }
```
Not so bad if we needed it I thought.

saudet:
It's a bit like the case of the `position()` method. It obviously can't return anything but itself, in the type of the current class, but we have to add a line in the subclass to cast it explicitly...

osialr:
Well not calling `super.close()` is bad form in the general case.   However, I will try to work something out to go along with #86 for C-struct based classes.  I think my usages can be solved adding this.

saudet:
I've given this some more thoughts. Since `Pointer.deallocate()` via `close()` is meant to execute a C++ destructor, we shouldn't be seeing exceptions thrown from there. Users that want to wrap something like `std::fstream::close()` that does throw exceptions, can still do so via some higher-level wrapper class in Java. So, let's make `Pointer` simpler to use. Thanks for bringing this up!

PR #87, opened 2016-03-11
Comment with first dimension of multidimensional array inside @Cast

osialr:
It's handy to see all dimensions of the input array even thought the cast needs to be `(*)`.

This will generate code such as `@Cast("const double(*"+/*[4]*/")[2]") DoubleBuffer quad`

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/tools/Parser.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Parser.java
@@ -825,7 +825,19 @@ public class Parser {
                     dimCast += "[" + dims[i] + "]";
                 }
             }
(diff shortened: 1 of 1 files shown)


saudet:
Good idea! We could even place them in C++ comments to make this more readable:
```java
@Cast("const double(*/*[4]*/)[2]") DoubleBuffer quad`
```
Or are Java comments better?

osialr:
Looking at them side-by-side, I prefer your method.  Then the `.cpp` will include the size as well and the unbalanced `+` goes away.  

I would put spaces in so it's easier to mentally parse.

```java
@Cast("const double(* /*[4]*/ )[2]") DoubleBuffer quad
```

PR #91, opened 2016-03-14
change arm build flags to work with arm6 as well as arm7

vb216:
Tested on pi1 and pi3, so looks good to support across multiple arms

### DIFF
--- a/src/main/resources/org/bytedeco/javacpp/properties/linux-armhf.properties
+++ b/src/main/resources/org/bytedeco/javacpp/properties/linux-armhf.properties
@@ -5,8 +5,8 @@ platform.includepath.prefix=-I
 platform.includepath=
 platform.compiler=g++
 platform.compiler.cpp11=-std=c++11
-platform.compiler.default=-march=armv7
-platform.compiler.fastfpu=-march=armv7-a -mfpu=neon -ffast-math
+platform.compiler.default=-march=armv6 -mfpu=vfp -mfloat-abi=hard
+platform.compiler.fastfpu=-march=armv7-a -mfpu=neon -mfloat-abi=hard -ffast-math
 platform.compiler.nodeprecated=-Wno-deprecated-declarations
 platform.compiler.output=-Wl,-rpath,$ORIGIN/ -Wall -O3 -fPIC -shared -s -o\u0020
 platform.linkpath.prefix=-L


saudet:
Great! Thanks for testing this out. We could just do it the way I did it for Android, for example:
```
platform.compiler.default=-march=armv6 -mfpu=vfp -mfloat-abi=hard
platform.compiler.fastfpu=-march=armv7-a -mfpu=neon -ffast-math
```
This way, users have the choice of "fasffpu" (ARMv7) when performance is critical and shouldn't bother using it on ARMv6, and of "default", when performance doesn't really matter that much, but portability does. What do you think?

BTW, should we also add " -mfloat-abi=hard" to fastfpu?

vb216:
Hm interesting idea.. I'm not sure how many people would go to the effort of setting up a cross build environment and tweak things tho? Thats why at first I thought lets go for portability, but seems to affect the opencv build quite a bit performance-wise. 

I've put a post out on pi forum [here](https://www.raspberrypi.org/forums/viewtopic.php?f=63&t=140244) asking if anyones tackled it, would seem strange to have to download chip specific builds (tho, I guess historically maybe people built locally for their chip - I remember things like MMX extensions back in the day)

Yeah I think the mfloat should be set to hard for both - seems soft float was only on v early Pi 1, all more recent distributions are hard float.

I guess lets see what the pi forum come back with

vb216:
Interestingly, did a bit more searching, and found loading the camera driver with "sudo modprobe bcm2835-v4l2 max_video_width=2592 max_video_height=1944" i.e. the extra parameters, made a difference in getting about 6 frames per second when just grabbing a frame, to 30+ ! So now the bottle neck looks more computation side. Adding back in code line by line, cvAbsDiff(grayImage, tempFrame, diffImage) drops the rate to 24fps and then cvThreshold(diffImage, diffImage, 64, 255, CV_THRESH_BINARY) drop to 10 fps. I guess they're more computationally heavy? Maybe why the compiler flags made a difference there?

deinhofer:
I think using computer vision on the pi only really makes sense if the performance exceeds a certain treshold and this was not the case before pi 3. So I think in this case performance is more important than portability.

vb216:
I guess it depends on the application? I've been using it for face recognition and doing fine with 3-4 frames a second on a pi2. Used it for marker detection too at that rate. Not ideal, but worked, so could be a shame to wipe out support on pi1/zeros (zero is an interesting platform too given the size). 

Personally, I'd add two builds in, an arm6 and arm7, and put some info in the wiki page explaining it along with info for how to setup a cross build environment (so its documented somewhere too!), maybe with a line about building locally on the target platform now (as Pi3 compile time isnt so bad). But, its not my project so I'm happy to go with whatever is preferred :)

saudet:
Well, we can't satisfy the needs of everyone. Most people are satisfied with a generic build that's fast enough and works on most devices. It's the same thing with x86. Most people don't worry about rebuilding their Linux distributions from scratch. Still, some people with more specialized needs can make builds for themselves. That's why the source code is available.

BTW, we have to differentiate between the JNI wrappers and the library itself. In almost no circumstances do we need to compile the wrappers with `-march=armv7-a -mfpu=neon -ffast-math` flag. OpenCV itself should probably be built that way, yes, but as long as the ABI doesn't change (hard float), we do not need to worry about compiling the JNI wrappers the same way.

saudet:
So, I'd say we go with ARMv6 builds, and not worry about "fastfpu" or anything, but still make it available so that users can make builds for themselves that way. If it turns out that a lot of people want ARMv7 builds, then we'll look into that again at that time. It's too early to say, and maintaining two sets of builds without more info is a waste of time, IMO. How does that sound @vb216?

vb216:
Yeah sounds good to me, and I agree, I think good to get a regular build that will work on all Pis included (I was using an old version 0.3 for ages as there was a precompiled Pi version available!) so that its taken up by that community more. 

I admit I'm not quite sure on the fastfpu / JNI wrapper side though - as it stands this pull request (+the javacpp-presets one) is setup for arm6, but I don't know if you want any more tweaks? The mvn snapshot deploy seemed to work fine too so hopefully we can get some feedback once people use it. 

I think there are still improvements possible on flags and using a newer compiler, but I've got a few other projects I need to move along first.

deinhofer:
Ok, I understand your point, there are certainly many use cases that are 
happy with the generic build.
We would like to experiment with face tracking on the pi which relies on 
a good frame rate. Hopefully, I will find some time in April, than I or 
my colleagues could help testing and tweaking the fast build for pi3. 
Please provide the documentation where to start...
Thanks!!

Am 17.03.2016 um 16:10 schrieb vb216:
>
> Yeah sounds good to me, and I agree, I think good to get a regular 
> build that will work on all Pis included (I was using an old version 
> 0.3 for ages as there was a precompiled Pi version available!) so that 
> its taken up by that community more.
>
> I admit I'm not quite sure on the fastfpu / JNI wrapper side though - 
> as it stands this pull request (+the javacpp-presets one) is setup for 
> arm6, but I don't know if you want any more tweaks? The mvn snapshot 
> deploy seemed to work fine too so hopefully we can get some feedback 
> once people use it.
>
> I think there are still improvements possible on flags and using a 
> newer compiler, but I've got a few other projects I need to move along 
> first.
>
> —
> You are receiving this because you commented.
> Reply to this email directly or view it on GitHub 
> <https://github.com/bytedeco/javacpp/pull/91#issuecomment-197925758>
>

-- 
Martin Deinhofer, MSc
Research and Development, International Projects

University of Applied Sciences Technikum Wien
Hoechstaedtplatz 6, 1200 Vienna, Austria, Europe
T: +43 1 333 40 77-297, F: +43 1 333 40 77-99 297
E: martin.deinhofer@technikum-wien.at
I: embsys.technikum-wien.at

saudet:
@vb216 Let's see, could you update the PR this way? I don't think we need more tweaks. Looks fine to me.
```
platform.compiler.default=-march=armv6 -mfpu=vfp -mfloat-abi=hard
platform.compiler.fastfpu=-march=armv7-a -mfpu=neon -mfloat-abi=hard -ffast-math
```
Basically, it'll target ARMv6 unless the user specifies the `compiler="fastfpu"` explicitly, which almost never happens, but it's good to keep these settings around. I wouldn't want to lose the info :)

@deinhofer Documentation, good point. @vb216 if you still have some time, please drop your notes on the wiki under a new "Linux (ARM)" section: https://github.com/bytedeco/javacpp-presets/wiki/Build-Environments

Thanks for everything guys!

vb216:
Sure, have changed the request in line with those flags now.
Added a Linux(arm) section to the wiki, hopefully there's enough info in there to help anyone pick it up

saudet:
Super!! Thanks a lot :)

deinhofer:
Thanks for the work!
Should the build actually also work for the older javacv versions like 0.10?
Then I will give it a try!!

vb216:
In theory.. I think so! If you have a working build setup for a given version, might be easiest just to paste/merge the linux-armhf sections in for the presets, along with the new definition in javacpp, but they're just build flags so I can't see why it wouldnt. If you hit an issue let me know, might have been something I came across along the way too

PR #94, opened 2016-04-30
fix for nullvalue becoming

ashleyj:
There are a few situations where nullValue is set to ```""``` and should remain ```""``` (empty strings for example), but the current code generates ```""""```, which causes compilation errors

eg.
```
@ByVal(nullValue = """") StringRef Name/*=*/);
```

This PR avoids the concatenation of the superfluous ```""```


saudet:
I'm wondering if those shouldn't be escaped instead to give us `nullValue = "\"\""`?

ashleyj:
Let me clean this up and create a new PR

saudet:
Ok, but you know, we can modify the branch the PR was created from: It will automatically get updated. No need to create new PRs.

PR #119, opened 2016-08-05
escape empty strings for nullValue

ashleyj:
Cleaned up version of previous PR. This will fix some problems with nullValue having a value of  """", which is invalid Java


saudet:
Thanks! Seems to be working fine this way.

PR #123, opened 2016-08-14
Add handling for OpenJDK to armhf build

vb216:
Tested on Pi1 with OpenJDK, seems to resolve the issue


saudet:
Great! Thanks for fixing this

PR #158, opened 2017-01-06
Skip duplicate 'using' statements

floybix:
Fixes bytedeco/javacpp-presets#353

I implemented as you suggested, and confirmed my preset now compiles without errors.


saudet:
Looks good! Thanks

PR #175, opened 2017-04-18
Update Loader.java

zhanhb:


saudet:
Good find! Thanks for the fix

PR #177, opened 2017-05-03
Removing unnecessary creation of a SecurityManager. Fixes #176.

### LINKED ISSUE #176: AccessControlException, javacpp trying to create a SecurityManager
Running with restricted privileges, I get s failure initialising due to an attempt to create a security manager:

```
java.lang.ExceptionInInitializerError
	at org.nd4j.linalg.cpu.nativecpu.ops.NativeOpExecutioner.<init>(NativeOpExecutioner.java:43)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at java.lang.Class.newInstance(Class.java:442)
	at org.nd4j.linalg.factory.Nd4j.initWithBackend(Nd4j.java:5629)
	at org.nd4j.linalg.factory.Nd4j.initContext(Nd4j.java:5553)
	at org.nd4j.linalg.factory.Nd4j.<clinit>(Nd4j.java:188)
	at Test.test(...)
Caused by: java.lang.RuntimeException: ND4J is probably missing dependencies. For more information, please refer to: http://nd4j.org/getstarted.html
	at org.nd4j.nativeblas.NativeOpsHolder.<init>(NativeOpsHolder.java:41)
	at org.nd4j.nativeblas.NativeOpsHolder.<clinit>(NativeOpsHolder.java:17)
	... 41 more
Caused by: java.lang.ExceptionInInitializerError
	at java.lang.Class.forName0(Native Method)
	at java.lang.Class.forName(Class.java:264)
	at org.nd4j.nativeblas.NativeOpsHolder.<init>(NativeOpsHolder.java:25)
	... 42 more
Caused by: java.security.AccessControlException: access denied ("java.lang.RuntimePermission" "createSecurityManager")
	at java.security.AccessControlContext.checkPermission(AccessControlContext.java:472)
	at java.security.AccessController.checkPermission(AccessController.java:884)
	at java.lang.SecurityManager.checkPermission(SecurityManager.java:549)
	at java.lang.SecurityManager.<init>(SecurityManager.java:299)
	at org.bytedeco.javacpp.Loader$1.<init>(Loader.java:262)
	at org.bytedeco.javacpp.Loader.getCallerClass(Loader.java:262)
	at org.bytedeco.javacpp.Loader.load(Loader.java:657)
	at org.nd4j.nativeblas.Nd4jCpu$NativeOps.<clinit>(Nd4jCpu.java:62)
	... 45 more
```

The code in the area already has some fallback logic which uses the modern API, `Thread.currentThread().getStackTrace()` - it seems like this fallback could just become the entire body of the method, rather than just being a workaround for Android.

trejkaz:
Previous code was constructing a SecurityManager and failing to catch
the SecurityException which this could throw.
Thread.currentThread().getStackTrace() is the modern API for looking
at the current stack, and much more likely to be permitted by the
security policy, and there was already code calling this newer method,
but for some reason it was only done as a fallback.

This change rewrites the method with the newer API as the only way.

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/Loader.java
+++ b/src/main/java/org/bytedeco/javacpp/Loader.java
@@ -259,34 +259,17 @@ public class Loader {
      * @return the Class found from the calling context, or {@code null} if not found
      */
     public static Class getCallerClass(int i) {
-        Class[] classContext = null;
         try {
-            classContext = new SecurityManager() {
-                @Override public Class[] getClassContext() {
-                    return super.getClassContext();
-                }
-            }.getClassContext();
-        } catch (NoSuchMethodError e) {
-            logger.error("No definition of this method : " + e.getMessage());
-        }
-        if (classContext != null) {
-            for (int j = 0; j < classContext.length; j++) {
-                if (classContext[j] == Loader.class) {
-                    return classContext[i+j];
+            StackTraceElement[] classNames = Thread.currentThread().getStackTrace();
+            for (int j = 0; j < classNames.length; j++) {
+                if (Class.forName(classNames[j].getClassName()) == Loader.class) {
+                    return Class.forName(classNames[i+j].getClassName());
                 }
             }
-        } else {
-            // SecurityManager.getClassContext() returns null on Android 4.0
-            try {
-                StackTraceElement[] classNames = Thread.currentThread().getStackTrace();
-                for (int j = 0; j < classNames.length; j++) {
-                    if (Class.forName(classNames[j].getClassName()) == Loader.class) {
-                        return Class.forName(classNames[i+j].getClassName());
-                    }
-                }
-            } catch (ClassNotFoundException e) {
-                logger.error("No definition for the class found : " + e.getMessage());
-            }
+        } catch (ClassNotFoundException e) {
+            logger.error("No definition for the class found : " + e.getMessage());
+        } catch (SecurityException e) {
+            logger.debug("Not permitted to look at stack: " + e.getMessage());
         }
         return null;
     }


saudet:
Using `SecurityManager` is faster:
http://stackoverflow.com/a/2924426/523744
I feel that should be a good enough reason to keep it...

trejkaz:
Hmm, is it called frequently enough that you have to worry about performance? I have to admit, I took a look and only saw stuff which looked like it was initialisation code.

saudet:
Initialization is slow, no need to make it any slower.

Is the code causing any problems for you?

trejkaz:
Issue #176.

trejkaz:
The irony, of course, is that this "faster" logic creating a security manager is actually slower for people actually running a security manager, since it has to fail to create that, and then fall back to some other behaviour.

Maybe it could check first, and not do the security manager hack if there is already one registered... to save the cost of exception handling. :/

saudet:
Hum, strange didn't get a notification from GitHub about that issue.

Ok, so what if we catch AccessControlException instead?

saudet:
Hum, strange didn't get a notification from GitHub about that issue.

Ok, so what if we catch AccessControlException instead?

saudet:
Or a more explicit check is great too, yes.

trejkaz:
You should catch SecurityException instead, since that's what the method docs say it could throw. There will still be exception handling overhead though.

saudet:
Could you update the PR so it does what we need? Thanks!

saudet:
I've added a catch block in commit 89e91f1. Let me know that it takes care of the issue! Thanks

trejkaz:
Cycling back to look at this now. Looks like that approach will fix the issue for me.

Issue #182, opened 2017-05-19
caching resource has race condition in case of multiple JVMs


mythly:
[extractResource](https://github.com/bytedeco/javacpp/blob/master/src/main/java/org/bytedeco/javacpp/Loader.java#L502) didn't consider the race condition, so the target resource file like `~/.javacpp/cache/opencv-3.2.0-1.3-linux-x86_64.jar/org/bytedeco/javacpp/linux-x86_64/libjniopencv_core.so` could be corrupted when multiple JVMs try to write the same file.

One can easily reproduce this by
```
rm -rf ~/.javacpp/cache/*
java <some jar using javacpp> & java <some jar using javacpp>
``` 
and see error likes
```
Caused by: java.lang.UnsatisfiedLinkError: /home/somebody/.javacpp/cache/something.jar/org/bytedeco/javacpp/linux-x86_64/libjniopencv_core.so: /home/somebody/.javacpp/cache/something.jar/org/bytedeco/javacpp/linux-x86_64/libjniopencv_core.so: ELF load command past end of file
```
or
```
Caused by: java.lang.UnsatisfiedLinkError: /home/somebody/.javacpp/cache/something.jar/org/bytedeco/javacpp/linux-x86_64/libjniopencv_core.so: /home/somebody/.javacpp/cache/something.jar/org/bytedeco/javacpp/linux-x86_64/libopencv_core.so.3.2: undefined symbol: _ZN2cv9AlgorithmD1Ev
```

Possible (but not good) solutions:
1. changing "org.bytedeco.javacpp.cachedir" of different JVMs to different folders
2. preparing all the cache before the first running of the application

Any better ideas?

saudet:
1. It used to work that way by default, and we can still make it work like that by setting the "org.bytedeco.javacpp.cachedir" system property to whatever we want:
http://bytedeco.org/javacpp/apidocs/org/bytedeco/javacpp/Loader.html#getCacheDir--
2. I'm already doing something like that by waiting that files are at least 1 second old:
https://github.com/bytedeco/javacpp/blob/master/src/main/java/org/bytedeco/javacpp/Loader.java#L426
If that isn't working anymore, it would be the first thing to investigate!

mythly:
"waiting that files are at least 1 second old" seems not working as both JVMs fall in the [earlier if branch](https://github.com/bytedeco/javacpp/blob/master/src/main/java/org/bytedeco/javacpp/Loader.java#L421).

The test code
```java
[...]

PR #188, opened 2017-06-05
Fix the race condition of `Loader.cacheResource()` (issue #182)

mythly:
See issue #182 for more details.

The fix was manually tested on only linux, as I cannot figure out the proper way to write an unit test, which would need two JVMs.


saudet:
Great, thanks! Locks might be expensive though. Maybe we should check the file size, time, and state twice, once before and after locking?

Also, let's name the file ".lock", if that's OK.

mythly:
Good point! I improved the implementation and ensured that locking happens only in the first running.

PR #198, opened 2017-08-21
Catch SecurityException when getting the cache directory.

trejkaz:
Logic in the area will already try the next directory.


saudet:
Thanks! LGTM

Issue #211, opened 2017-12-06
JNI_OnLoad_library used before it is declared


johanvos:
When compiling a static library, the JNI_OnLoad_library pattern is used. However, the generated file contains a call to JNI_OnLoad_library before it is used.

PR #212, opened 2017-12-06
Fix undefined declaration in case of static libs

johanvos:
This is a fix for #211 
In case JNI_OnLoad_library is used, we need to declare it before it is called.


saudet:
Thanks! Keep them coming :)

PR #221, opened 2017-12-14
add includepath for atomic on ios

johanvos:
Make sure <atomic> can be found when compiling for ios  (at least x86_64)


saudet:
Does that work with the version of clang from Xcode?? And what about other architectures?

saudet:
If this is to get libnd4j to build, we should fix this over there...

johanvos:
That is with 
* Apple LLVM version 9.0.0 (clang-900.0.39.2)
* Xcode 9.2

This is indeed to build nd4j  (libnd4j//blas/Environment.h includes <atomic>)

saudet:
Xcode doesn't officially support that. BTW, it also doesn't support OpenMP and that's why we use GCC on other platforms. Would there be a way to use GCC with iOS too?

saudet:
Let's make this modification in ND4J, see pull https://github.com/deeplearning4j/nd4j/pull/2538.

PR #226, opened 2018-01-25
Additional catch block for Loader

AlexDBlack:


saudet:
Thanks for the fix! I've adjusted other places where symbolic links might get created. Could you test and make sure that this still works?

AlexDBlack:
Apologies for the delay - I was finally able to test this again, and confirm that it fixes the originally encountered issue.

PR #244, opened 2018-07-11
Fix arm/arm64 clang properties

tmm1:
This fixes build failures when I tried to switch to the new clang compiler


src/main/resources/org/bytedeco/javacpp/properties/android-arm64-clang.properties
```diff
@@ -22,7 +22,7 @@ platform.linkpath.prefix=-L
 platform.linkpath=sources/cxx-stl/llvm-libc++/libs/arm64-v8a/
 platform.link.prefix=-l
 platform.link.suffix=
-platform.link=c#:m#:z#:dl#:android_support#:c++abi#:c++_static#:log#
+platform.link=c#:m#:z#:dl#:c++abi#:c++_static#:log#
```
saudet: I'm pretty sure `android_support` is required here. What kind of error are you getting?

tmm1: Linker error, library not found. From what I saw it's not required on arm64

https://github.com/googlesamples/android-ndk/issues/530#issuecomment-390749822

I verified this PR on actual devices of all three platforms.

saudet: Yeah, it looks like it was removed from r17b. I wonder if we can get away without it even with r15c or r16b, let me check that...

saudet: Nope, we need to link with `android_support` in the case of r15c or r16b... hum...

saudet: Well, let's take it out I guess since r17 is probably going to be a starting point for many using Clang. We can always readd it later as part of a `BuildEnabled` class.

Given that, let's also remove it from `android-x86_64-clang.properties` as well.

tmm1: > Given that, let's also remove it from android-x86_64-clang.properties as well.

Done.

tmm1: Just ran into the same thing but in the x86 profile. I will make a new PR

PR #246, opened 2018-07-28
Fix race condition of loadedLibraries

louxiu:
We should put lib to loadedLibraries only after System.load(lib)
success, or loadedLibraries.get(lib) will get it in other thread
before we remove it after System.load(lib) failed.

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/Loader.java
+++ b/src/main/java/org/bytedeco/javacpp/Loader.java
@@ -1201,12 +1201,11 @@ public class Loader {
                         if (logger.isDebugEnabled()) {
                             logger.debug("Loading " + filename2);
                         }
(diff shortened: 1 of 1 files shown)


saudet:
The reason we can't do that is because `System.load()` calls `JNI_Load()`, which might start loading other classes that may include the class we're already trying to load, as part of some cycle. We need a way to detect that we're already loading a given class. What issues are you encountering with the current scheme?

saudet:
I see, a race condition could occur in there if we're not loading from a `static { }` block or something. Those are basically `synchronized`, maybe we should make `Loader.loadLibrary()` synchronized as well?

saudet:
When we link against shared libraries, yes, we need to try to load them. But we can disable that behavior by suffixing them with `#`, which is used to rename library files on load, but if no name is given, they just don't get loaded.

Issue #247, opened 2018-07-31
link static lib to dynamic lib on linux


louxiu:
I also come across this problem [libjni*.so not containing all symbols from .a](https://groups.google.com/forum/#!topic/javacpp-project/CorLDiMjATM) 

>        --whole-archive
>              For  each archive mentioned on the command line af-
>              ter the --whole-archive option, include  every  ob-
>              ject  file  in the archive in the link, rather than
>              searching  the  archive  for  the  required  object
>              files.   This  is  normally used to turn an archive
>              file into a shared library, forcing every object to
>              be included in the resulting shared library.

If I understand the option correctly, when the `--whole-archive` is not added, only the used symbol will be add to the final dynamic lib and other will be leaved undefined. But why there is no error on the mac(without `-all_load` or  `-force_load` ).

saudet:
It's probably a limitation of the version of GCC you're using...

saudet:
If you send a pull request it looks like something I'd merge though :)

louxiu:
Glad to do that. 😃  I will do more study on this as I am not fully understand it.

PR #250, opened 2018-08-12
ADD linkbefore and linkafter option

louxiu:
To have chance to add more linker options. #247


saudet:
Allowing to set compiler options directly in the Java files makes it harder to set properties for multiple platforms. Also, it's just occurred to me that we can probably already use `platform.link.prefix` and `platform.link.suffix` to solve this issue cleaning, for example:
```
platform.link.prefix=-Wl,--whole-archive -l
platform.link.suffix= -Wl,--no-whole-archive
```
Should work if we update the `Builder` a bit to split the properties into multiple options at the white spaces, like it's already doing for `platform.compiler`. Could you give that a try?

louxiu:
Sorry for late reply. It's a clever way. 😃  I resubmit a commit. And should we change the property file(`linux-x86_64.properties` etc) or there is another way to override the default value of `platform.link.prefix` and `platform.link.suffix`?

saudet:
Yes, that's what the "properties" in the command line option and in the Mojo configuration are for. I wonder if we can make this the default though?

We want to merge the last prefix with the library name, but only when it doesn't end with a space, and similarly for the first suffix. Are you sure your changes do that?

louxiu:
Ok, I know what's your preferred way. Please check it again.

saudet:
No, that doesn't split all spaces. We can do like it was before, but consider the parts after the first "word" in the suffix as independent options, and similar for prefix.

louxiu:
@saudet done

saudet:
Looking better, almost there, but with the following properties when there are no libraries to link (like with the unit tests that you can try by running `mvn test`)  it appends a `-Wl,--whole-archive` without a `-Wl,--no-whole-archive`. Moreover, when we do specify a library to link, the suffix gets appended to the library name like `-lfoo-Wl,--no-whole-archive`, when that should be 2 separate options.
```
platform.link.prefix=-Wl,--whole-archive -l
platform.link.suffix= -Wl,--no-whole-archive
```

saudet:
Ah, yes, good catch! That makes it a bit harder to use, but oh well. Using a backlash to escape like this also works:
```
platform.link.prefix=-Wl,--whole-archive -l
platform.link.suffix=\ -Wl,--no-whole-archive
```
Anyway, now it looks like the whole prefix gets appended to the library name in only 1 option, including the space, like this: `g++ ... '-Wl,--whole-archive -lfoo' ...`. Those need to be 2 options.

louxiu:
No, if I test correctly. The latest commit has splitted `-Wl,--whole-archive -l` as 2 options.

saudet:
You're right, I was testing on the wrong branch, sorry. Seems to work correctly, but it's stripping the options in between each library. I'm not sure it always makes sense to do that? It wouldn't hurt to leave them at least... What do you think?

louxiu:
I think I got what you mean finally. 
It should be 

 `-Wl,--xxx -Wl,--whole-archive -lfoo -Wl,--no-whole-archive -Wl,--whole-archive -lbar -Wl,--no-whole-archive` 👍 
 `-Wl,--xxx -Wl,--whole-archive -lfoo -lbar -Wl,--no-whole-archive` 👎 
 `'-Wl,--xxx -Wl,--whole-archive' -lfoo -Wl,--no-whole-archive '-Wl,--xxx -Wl,--whole-archive' -lbar -Wl,--no-whole-archive` 👎 

Sorry for my bad English. Is it right?

saudet:
It would actually be like this though:
```
-Wl,--xxx -Wl,--whole-archive -lfoo -Wl,--no-whole-archive -Wl,--xxx -Wl,--whole-archive -lbar -Wl,--no-whole-archive
```

louxiu:
Yes, I prefer that way too. It keeps meaning of `link.prefix` and `link.suffix`.  I misunderstand your comment. 😢 

> We want to merge the last prefix with the library name, but only when it doesn't end with a space, and similarly for the first suffix.

PR #252, opened 2018-08-22
[WIP] Add support for Java's module system

SamCarlberg:
# Overview
Splits the project into two subprojects: `core` and `maven-plugin`. The plugin has dependencies that cannot be modularized, which prevents downstream applications from using `jlink` to create native runtime images. The core project has no such dependencies, and can be used with `jlink` 

`Loader` now looks for property files with absolute paths (`/properties/...`) instead of relative ones (`properties/...`)

Bumps the version to 2.0.0-SNAPSHOT. This might not be necessary, since the artifact name has been changed from `javacpp` to `javacpp-core` and `javacpp-maven-plugin` for the two projects

How the projects in [javacpp-presets](https://github.com/bytedeco/javacpp-presets) will be modularized remains TBD. However, they will also need to be modularized (and moved to different packages) to avoid exporting the `org.bytedco.javacpp` package from multiple modules

# TODO
- [ ] Ensure that applications built on Java 9 and above still work; in particular, make sure `Loader` will still be able to resolve and load native binaries bundled inside the JAR
- [x] Release a multi-release JAR that can be used by both Java 7 and Java 9 consumers
- [ ] Update README with instructions for the new artifacts


saudet:
Looks like multi-release JARs are still pretty buggy after all:
http://mail.openjdk.java.net/pipermail/jigsaw-dev/2018-September/013932.html
That doesn't fill me with confidence...

/cc @HGuillemet

HGuillemet:
I'm not familiar with multi-release jar and don't know exactly why you would need it, but modules are the future, and I think you should consider doing the needed refactor ASAP and before you have too many users having worked with the current structure.
Different jars must export different packages. Would it be possible to have all classes specific to the opencv presets lie  in org.bytedeco.opencv package, for instance, or org.bytedeco.javacpp.presets.opencv ?

saudet:
Yes, but classes like `opencv_core` are already pretty much packages, and it would it make it harder to link between native libraries since the concept of "package" doesn't exist at the native level. In my opinion, modules still have a long way to go before becoming accepted by the industry, there is no hurry, but let's make progress, please try stuff out! Thanks for your interest

saudet:
To work around that, we could start using proper Java packages. We could still have, for example, `org.bytedeco.javacpp.opencv_core.Mat` for classes, but for global functions and variables, we would have to put them in separate artificial classes like `org.bytedeco.javacpp.opencv_imgcodecs.global.imread()`. By adjusting the imports as follows, it shouldn't be too hard for users to "upgrade" though:
```java
import org.bytedeco.javacpp.opencv_core.*;
import org.bytedeco.javacpp.opencv_imgcodecs.*;
import static org.bytedeco.javacpp.opencv_core.global.*;
import static org.bytedeco.javacpp.opencv_imgcodecs.global.*;
```
The comments in the header files might also get harder to relate that way, but on the plus side, this structure would be more Java friendly in general, not just for modules. What do you think?

SamCarlberg:
Would `opencv_core` in this scenario be a package or a class?

saudet:
A Java package, that's what we need to add, right?

SamCarlberg:
Just wanted to make sure. Does the generator support generating multiple classes into a single package? I thought it could only generate everything into a single file

saudet:
It doesn't currently because it's easier to map header files and native libraries all to the same "namespace", but it could be modified that way.

HGuillemet:
Splitting the huge opencv_core class  would definitely be easier to use and understand. But it'd look more logical if the package was org.bytedeco.opencv, core being either a class or a package inside it. That would fit the names in the C++ structure. If it helps, core could even be both a class and a package in this layout, but that's probably a bad idea (obscured declarations).

HGuillemet:
ok, I was looking at source names. 
So you either can consider each library as a different preset, and using packages 
org.bytedeco.opencv_core, etc...
Possibly shipping these different "presets" in a single archive.
or define a "preset" package level : org.bytedeco.opencv, then a library "sublevel": org.bytedeco.opencv.opencv_core, with would look nicer if you can configure the preset to get rid of the opencv_ prefix.

saudet:
In any case, we can experiment with that easily once we add the ability to target a package, so let's get this done first. Please let me know if you're interested in contributing, I'll help with any issues that you encounter. Thanks for your interest!

HGuillemet:
I'm still feeling a newbie about JavaCPP and would probably take 10x the time you need to do the same, but if you can think of a precise task you'd like me to do and can give me some directions, I'll try to help.

saudet:
I've added support to target packages, so we can start experimenting with the presets! We just need to modify the presets with `@Properties(..., target = "...", global = "SomeClassName")`, and the name in `target` will be considered as the name of the Java package where to output the global class as well as all the other classes. We also need to adjust some `import` statements here and there, and the `pom.xml` files a bit, something like this for the parent one and similarly for some of the submodules:
```diff
diff --git a/pom.xml b/pom.xml
index 641b5cb1c..d6793a5d7 100644
--- a/pom.xml
+++ b/pom.xml
@@ -152,7 +152,7 @@
             <id>default-compile</id>
             <configuration>
               <includes>
-                <include>org/bytedeco/javacpp/*.java</include>
+                <include>org/bytedeco/javacpp/**/*.java</include>
               </includes>
             </configuration>
           </execution>
@@ -282,7 +282,7 @@
             </goals>
             <configuration>
               <skip>${javacpp.compiler.skip}</skip>
-              <classOrPackageName>org.bytedeco.javacpp.*</classOrPackageName>
+              <classOrPackageName>org.bytedeco.javacpp.**</classOrPackageName>
               <copyLibs>true</copyLibs>
               <copyResources>true</copyResources>
             </configuration>
@@ -317,16 +317,16 @@
                 </manifestEntries>
               </archive>
               <includes>
-                <include>org/bytedeco/javacpp/*</include>
+                <include>org/bytedeco/javacpp/**</include>
                 <include>org/bytedeco/javacpp/helper/*</include>
                 <include>org/bytedeco/javacpp/presets/*</include>
               </includes>
               <excludes>
                 <exclude>org/bytedeco/javacpp/include/</exclude>
-                <exclude>org/bytedeco/javacpp/linux-*/</exclude>
-                <exclude>org/bytedeco/javacpp/macosx-*/</exclude>
-                <exclude>org/bytedeco/javacpp/windows-*/</exclude>
-                <exclude>org/bytedeco/javacpp/${javacpp.platform}${javacpp.platform.extension}/</exclude>
+                <exclude>org/bytedeco/javacpp/**/linux-*/</exclude>
+                <exclude>org/bytedeco/javacpp/**/macosx-*/</exclude>
+                <exclude>org/bytedeco/javacpp/**/windows-*/</exclude>
+                <exclude>org/bytedeco/javacpp/**/${javacpp.platform}${javacpp.platform.extension}/</exclude>
               </excludes>
             </configuration>
           </execution>
@@ -342,7 +342,7 @@
               <includes>
                 <include>${javacpp.platform.library.path}/</include>
                 <include>${javacpp.platform}${javacpp.platform.extension}/</include>
-                <include>org/bytedeco/javacpp/${javacpp.platform}${javacpp.platform.extension}/</include>
+                <include>org/bytedeco/javacpp/**/${javacpp.platform}${javacpp.platform.extension}/</include>
               </includes>
               <excludes>
                 <exclude>org/bytedeco/javacpp/windows-*/*.exp</exclude>
```

HGuillemet:
I have started to use the new package-targetting feature on the opencv presets to progress towards molularization.

1. the org.bytedeco.javacpp.presets package was moved to org.bytedeco.javacpp.opencv.presets
2. target changed for each library to org.bytedeco.javacpp.opencv.imgproc and the like
3. added for each library global="ImgProc" and the like
4. for consistency, helpers class were refactored: all static inner classes included in each helper class were moved as standalone classes to target package.
5. if the helper class didn't contain any methods (only static inner classes), it was removed. Else it was renamed "Helper", turned package-private and moved to target package.
6. for some reason, Mat constructors could access the protected fields limit and position in Pointer before the refactoring and not after, so limit and position replaced by limit() and position().
7. in order to prevent helper classes to mix with generated classes, classes are now generated into target/generated-sources directory

Result in this [commit](https://github.com/HGuillemet/javacpp-presets/commit/1d59ec4b7c102b9bf63bb79b6f287ffbc730a4d0).

Java part compiles, but not the JNI part: 
jnijavacpp.cpp and org/bytedeco/javacpp/opencv/video/jniPointer.cpp are generated but contain reference go CvMat and CvArr, which triggers: 
`error: 'CvArr' was not declared in this scope`

There must be some mistakes in the paths I changed in one of the pom.xml. It looks like the JNI generator cannot access the classes moved to generated-sources, or the helper classes.

Any idea about what's wrong ?

saudet:
Thanks for testing this out @HGuillemet! Before we get too far into the weeds though, I'd like to see if we couldn't settle something that has been bothering, the names of the "global" classes. You've used `Core`, `Calib3D`, etc, which is mostly in line with the official Java API, so potentially creating conflicts. Following this pattern, in the case of FFmpeg, we would get `AVUtil`, `AVCodec`, `AVFormat`, etc, but that conflicts with a few existing structs like `AVCodec`, `AVFilter`, etc. For TensorFlow, I suppose `TensorFlow` could work, but again creating potential conflicts with the official Java API. So, this naming scheme doesn't seem optimal to me.

I've been thinking we could just name them all the same like `global`, and that works alright with `import static` for the most part, but not with normal `import`, and the links with Javadoc end up all having the same name.

We could keep the names lowercase as another option, that would solve these issues, but then we'd end up with fully qualified names like `org.bytedeco.javacpp.avutil.avutil`, not so great either, but at least it would be consistent.

Would you guys have any other ideas?

SamCarlberg:
The globals class for each module would be better named `Globals` or `${module}Globals` (eg `CoreGlobals`) to better convey what it's for. This also sidesteps class name collisions.

I think more classes should get pulled out of the globals class, too - `Mat` should be a top-level class in the `opencv.core` package, not a nested class of `opencv.core.CoreGlobals`. Not sure how feasible this is with the current state of the Generator.

saudet:
Something like `CoreGlobal`, `AVFilterGlobal`, `TensorFlowGlobal`, etc sounds interesting, but I'm not sure that really clarifies things because those are not names that are part of the C++ API anyway. It makes them look like API classes, even though they are not.

In any case, with the changes in commit https://github.com/bytedeco/javacpp/commit/f89f1c370617105f940a4e7a0eabbbcb491939dd it's now possible to target a package for classes such as `Mat`, so those don't need to end up in a "global class" anymore, no.

I'm tempted to go with the "double name", plus it would make it easier to upgrade existing code, for example:
```java
import org.bytedeco.javacpp.opencv_imgcodecs;
import static org.bytedeco.javacpp.opencv_core.*;

Mat image = opencv_imgcodecs.imread("image.png");
```
In this case, we would only need to update the imports, and it's also clear that `opencv_core` and `opencv_imgcodecs` refer to the corresponding native library, both the package and the global class:
```java
import org.bytedeco.javacpp.opencv_core.*;
import org.bytedeco.javacpp.opencv_imgcodecs.*;

```

saudet:
BTW, if this is just to get it working with `jlink`, I don't think we need to change anything really. The dependencies for the Maven plugin are optional, and AFAIK anything else can be shaded into an uber JAR before feeding it to `jlink`.

HGuillemet:
Sorry @saudet, I read to fast your comment: we agree on your example.
But this raise another question: 
org.bytedeco.javacp.opencv_core or org.bytedeco.javacpp.opencv.core ?

saudet:
The idea with `opencv_core`, etc is to use the same names as the native libraries. Since JavaCPP's goal is to bridge with C++, we can't expect to follow all Java conventions to come up with an interface most consistent with the original API. Whenever we see a C++ example on some blog somewhere linking with say `opencv_calib3d`, I think it helps to use the same names somehow. Besides, OpenCV 4.0 is not changing the library names, so it looks like they're here to stay.

Another option would be to go monolithic. Most libraries ship that way by default these days, like TensorFlow, and the ones that don't often end up with spaghetti dependencies like OpenCV or FFmpeg, but there are some for which it makes sense like CUDA, not sure I'd see that one as monolithic. It wouldn't be inconsistent to leave it like it is though, the concept of "fat library" does exist in native land.

HGuillemet:
If there are no use cases where the intermediate package could host something interesting, keeping the lib name sounds ok, with the advantage of limiting code changes of existing users.
I don't really see the interest for a monolithic opencv though. Many people uses a small subset of the libraries.

SamCarlberg:
If everything is in the same project, then the `module-info` file will have to specify dependencies on the Maven plugin APIs, aether, etc.

The plugin module dependencies wouldn't have to be passed on to library consumers, since module dependencies can be optional, but the real problem is that the Maven API libraries have multiple JAR files (which each get resolved to a different automatic module) exporting classes from the same package. I don't remember which libraries have this problem, but it's along the lines of:

| Library | Module | Class
---|---|---
`a-1.2.3.jar` | `a-123` | `com.example.ClassA`
`b-4.5.6.jar` | `b-456` | `com.example.ClassB`

This could be worked around manually, but I'm not sure how it could be done with Maven. I haven't touched that side of the module system

saudet:
If we don't specify any `requires`, the compiler seems to ignore anything in the class path. I tried with this `module-info.java`:
```java
module org.bytedeco.javacpp {
  requires jdk.unsupported;
  exports org.bytedeco.javacpp;
  exports org.bytedeco.javacpp.annotation;
  exports org.bytedeco.javacpp.indexer;
  exports org.bytedeco.javacpp.tools;
}
```
And with these commands, I'm getting a usable `javacpp.jar` and a 12 MB `javacppapp.tar.xz` archive:
```bash
export CLASSPATH=`mvn dependency:build-classpath | grep -v "[INFO|WARNING]"`
javac -source 7 -target 7 -d target/classes `find src/main/java -iname '*.java'`
cp -a target/classes target/classes-9
javac -source 9 -target 9 -d target/classes-9 module-info.java
rm -Rf target/classes-9/org
jar --create --file javacpp.jar -C target/classes . --release 9 -C target/classes-9 .
jlink --module-path . --add-modules org.bytedeco.javacpp --output javacppapp
tar -cJvf javacppapp.tar.xz javacppapp/
```
Now we just need to figure out how to make Maven do that...

HGuillemet:
> Another thing I've just realized, with nested classes, we can more easily work around conflicts with more Java-like APIs, such as the official API of OpenCV:
> 
> ```java
> import org.opencv.core.Mat;
> import org.bytedeco.javacpp.opencv_core;
> 
> Mat mat1 = new Mat();
> opencv_core.Mat mat2 = new opencv_core.Mat();
> ```
> 
> By putting classes naked in a package we lose that advantage, so it's not clear-cut to me anymore that it would be best to change this. 

OTOH, we can argue that it'd be better for the user that JavaCPP mimics the official API, so that [s]he is not too lost when switching to JavaCPP after realizing the limitations of the official API. And I see the official API uses the exact package structure similar to the one I started to implement, that is *.opencv.core.Core, *opencv.core.Mat, etc... (that was not intended :)). I guess you'll have more users making this switch than users switching from C++ OpenCV to JavaCPP OpenCV.

I haven't come up with a use case when you need to use both JavaCPP and the official API. Is there really any ? I even wonder why you ship the official API with JavaCPP-OpenCV. The 2 API are not interoperable, are they ? Even if that might happen, fully qualified class names can always be used and I don't think that should orient the choice for your structure.

Still from a user point of view, I consider the javadoc of my build less confusing that the inner-class version (easier to navigate through). 
BTW, one advantage of the JPMS is to limit the packages you export. If you keep the present structure, It could be interesting to not export org.bytedeco.javacpp.helper and org.bytedeco.javacpp.presets.
 
> For JPMS support, we could always add/rename a package based on the name of the module, so something like `org.bytedeco.opencv.opencv_core`, `org.bytedeco.ffmpeg.avutil`, `org.bytedeco.tensorflow.tensorflow`, etc.

Yes, adopting a org.bytedeco.<preset/module>.<library> hierachy seem simple and logical.
I don't know if bytedeco is meant to deliver something else than JavaCPP and if you can strip the javacpp segment. If you come up with a preset where there is a header file shared by multiple libraries, you could put it in org.bytedeco.<preset/module>.

saudet:
@HGuillemet All that to say, it would be great if you could be a sort of "ambassador" to OpenCV, taking care of relations with them and eventually have them use JavaCPP in the `org.opencv` namespace. Is it something you'd like to do? And the idea would be to have someone else for TensorFlow, another for FFmpeg, Panama, etc. That's basically how Fedora works!

As for the refactoring of the presets, I think I'd settle on the following, putting generated sources in `src` (though other than `main`) instead of `target` so they can still be checked in the repository and be made searchable:
```
src/main/java/org/bytedeco/javacpp/opencv_core/opencv_core_helper.java
src/main/java/org/bytedeco/javacpp/opencv_core/opencv_core_presets.java
src/main/java/org/bytedeco/javacpp/opencv_core/AbstractMat.java
src/gen/java/org/bytedeco/javacpp/opencv_core/opencv_core.java
src/gen/java/org/bytedeco/javacpp/opencv_core/Mat.java
...

src/main/java/org/bytedeco/javacpp/avcodec/avcodec_helper.java
src/main/java/org/bytedeco/javacpp/avcodec/avcodec_presets.java
src/gen/java/org/bytedeco/javacpp/avcodec/avcodec.java
src/gen/java/org/bytedeco/javacpp/avcodec/AVCodec.java
...

src/main/java/org/bytedeco/javacpp/tensorflow/tensorflow_helper.java
src/main/java/org/bytedeco/javacpp/tensorflow/tensorflow_presets.java
src/gen/java/org/bytedeco/javacpp/tensorflow/tensorflow.java
samples/src/main/java/org/bytedeco/javacpp/tensorflow/samples/...
...
```
I think that would provide the path with lowest friction for users to upgrade, while still providing a good Java experience. We could name the global classes like `opencv_core_global`, but I'm not sure the longer names are worth it since those are referenced often. Including "helper" and "presets" as classes in the package would allow us to make them private too, but I don't think we can do that anyway because it's not possible to use "inherited" static methods from, for example, Scala, and JavaCPP needs to be able to access the presets from outside the package. I also think it makes sense at this point to come up with a new naming convention. Java classes are typically named with camel case, but following a different convention for these classes that map to native libraries has 2 advantages: It prevents name clashes with actual API classes, and allows to more easily consult C++ references and port examples written in C/C++. Plus it makes it easier to decide on a good name, almost always by just picking the name of the native library. :) What do you guys think? @SamCarlberg @HGuillemet 

About the "urgency", with the pending breaking releases of OpenCV 4.0.0 and TensorFlow 2.0, now may be a good time to do this kind of breaking changes for JavaCPP as well after all!

HGuillemet:
> Correct, they'd have to be in the same package... BTW, do you have any current needs with regards to JPMS? It doesn't sound like something that needs to be addressed right away, for example, we can always use `jlink` via an uber JAR, but let me know if you have any more urgent needs. In any case, let's keep the discussion going and thanks for your help!!

I'm just starting the developments of my opencv/javacpp/javafx application, so no personal need for packaging an optimized jre before some months, and I could workaround with an uberjar indeed. But this move to modularisation should be done and since it implies some change in JavaCPP API, the sooner the better...



> it would be great if you could be a sort of "ambassador" to OpenCV, taking care of relations with them and eventually have them use JavaCPP in the `org.opencv` namespace. Is it something you'd like to do?

I'll try to start a discussion there about JavaCPP.

> As for the refactoring of the presets, I think I'd settle on the following, putting generated sources in `src` (though other than `main`) instead of `target` so they can still be checked in the repository and be made searchable

It seems a bit weird to add to the git repository files that are automatically generated during the maven build. But if you think it's useful to be able to browse them from the repository web interface... why not.

>  `src/gen/java/org/bytedeco/javacpp/opencv_core/Mat.java`

So targetting packages instead of class ? I agree about this. In fact I don't really see any significant  advantage of the big inner class.
About stripping the <preset/module> level in the package hierary, I have no opinion : shorter package is nice, limiting the necessary changes for existing users is a good point too. I don't know if you might have a use to define a class common to all libraries of a preset. Something like a common header file may be rare enough, and even if that happens, you could always define an "artificial" package, not corresponding to a library, just for hosting the common class, like org/bytedeco/javacpp/opencv/Common.java. Devoting a level package for such rare case is probably not a good idea. So I tend to agree with this choice too, unless there is another use for a preset level package I'm not aware of.

> `src/gen/java/org/bytedeco/javacpp/opencv_core/opencv_core.java`

You points about keeping opencv_core naming convention for the global class sounds logical too.

> `src/main/java/org/bytedeco/javacpp/opencv_core/opencv_core_helper.java`
> `src/main/java/org/bytedeco/javacpp/opencv_core/opencv_core_presets.java`
>Including "helper" and "presets" as classes in the package would allow us to make them private too, but I don't think we can do that anyway because it's not possible to use "inherited" static methods from, for example, Scala, and JavaCPP needs to be able to access the presets from outside the package.

I know nothing about Scala, but if JavaCPP must access the preset class then there is no point of keeping them in the main package. I'd suggest:

`src/main/java/org/bytedeco/javacpp/opencv_core/presets/Preset.java`

making the Preset class public, and specifying a qualified export in the module-info.java, that is:

`export org.bytedeco.javacpp.opencv_core.presets to org.bytedeco.javacpp`

The case of the helper class is a bit different. Having opencv_core_helper public in the main package doesn't seem a big deal if working around the Scala limitation is an issue.

> About the "urgency", with the pending breaking releases of OpenCV 4.0.0 and TensorFlow 2.0, now may be a good time to do this kind of breaking changes for JavaCPP as well after all!

 :+1:

saudet:
I've created `jpms` branches in this repo and in the presets repo as version 1.5-SNAPSHOT to start testing things out. @HGuillemet has already started to work on OpenCV in pull https://github.com/bytedeco/javacpp-presets/pull/649.

@SamCarlberg If there is anything else that should be done in this repository for JPMS, please update this pull request and let's merge it! If you foresee breaking changes, also change the target branch to `jpms`:  https://help.github.com/articles/changing-the-base-branch-of-a-pull-request/ Thanks!

saudet:
BTW, Stephen Colebourne @jodastephen posted some useful information to help us understand what we might be able to do for the `Loader` here:
http://mail.openjdk.java.net/pipermail/jigsaw-dev/2018-July/013872.html

SamCarlberg:
Sorry about the radio silence.

How did you add a single `module-info` file while keeping the library  and Maven plugin in the same project?

As for the `Loader`, I'm not familiar enough with that class to give an opinion.

saudet:
@SamCarlberg No problem, we all don't have enough time. :) As I've shown above in pull https://github.com/bytedeco/javacpp/pull/252#issuecomment-437248927, the compiler from Java 9 ~ 11 works the same as Java 7 ~ 8 when targeting those versions for all the classes, and it doesn't complain about anything when compiling *only* `module-info.java`, so that's what we're doing. After figuring out a few more things (see pull #276), `Loader` also seems to be working well without breaking anything.

saudet:
Maven plugins use the class path, not the module path, they will probably
never support modules either, and non-Maven users won't use the plugin
classes, so there is no issue that I can see of. Could you elaborate?

SamCarlberg:
The current implementation on `master` seems to perform correctly. I can use the javacpp module in a modular application with no compile or runtime issues. As expected, there are problems when using a presets package like OpenCV because of the split package between the OpenCV presets module and the javacpp module.

Splitting the core libraries and the maven plugin into separate projects would be a good logical distinction, but since the plugin classes only add a few KB to the final size of the JAR I don't think it's a big deal.

With that said, I think this PR has been obsoleted by 7bff797bb91d75b04890598609ff014fa7adae7f. The presets will still need to be moved to unique packages, which I think would be better discussed on gitter or on an issue in the javacpp-presets repo than on this PR

saudet:
Sounds good to me, thanks for the input! (Personally I dislike libraries that split everything into 5 KB modules...) @HGuillemet has started for the presets with https://github.com/bytedeco/javacpp-presets/pull/649. How should we proceed? I could apply it to the `jpms` branch right away if that makes it easier to collaborate.

SamCarlberg:
There's an argument to be made about splitting the Maven plugin, a completely independent codebase that no downstream consumer of the `javacpp` library will ever use, out of the same project (and indeed the same _package_) as the code library. But since it doesn't prevent the modularization of the library, it's orthogonal to the modularization of the project.

If @HGuillemet could summarize the changes made to the OpenCV preset, I'm all for merging it and using it as a template for the other presets. But I'm not going to grep through a 70kloc diff to see the changes that need to be made across all the presets :)

saudet:
There is an argument against it as well: It create more stuff to maintain and takes more time to reply to people who don't understand why things don't work with only 1 module, etc. When you guys are able to take some of the load off, then we'll talk about increasing time required for maintenance and support.

saudet:
@SamCarlberg BTW, before we go on modifying all the other presets, we still need to figure out how to generate and compile a `module-info.java` for each classifier artifact separately. @HGuillemet has been thinking of trying out https://github.com/moditect/moditect for this. Feel free to try something else though.

HGuillemet:
Concerning the split, there is also the parser and the generator that could be shipped with the maven plugin, so 1 jar for building presets, and 1 jar to run them. But I agree  that's not the priority, if ever it's a good idea.

Concerning the modularization of a preset, I'll post a recipe, but we'll need to decide something for the module-info of the presets, since the chosen solution would change the recipe.

The problem with this module-info is that its content depends on the system and arch. So the easiest would be to generate it inline in the pom using the `${os.arch}` property and similar. That is possible with the moditect module. But complicated in the current pom because the same pom is used to generate 2 jars : the arch-independent and the arch-dependent jars. That's why I suggested Samuel to create a maven sub-module to generate the arch-dependent jars. Who can try this change ? I could, but not before next week.

saudet:
We can create multiple JAR files from the same `pom.xml` by calling the JAR plugin multiple times, so I'm pretty sure we can also create multiple `module-info.java` file by calling the ModiTect plugin.

saudet:
BTW, we could have a version of JavaCPP that is split into 2 or 3 modules in the `jpms` branch and test it out that way. I would be fine with that!

SamCarlberg:
@saudet Good to hear!

I'm going to close this PR since it seems that modularity has already been added. I'd be happy to keep this thread going for continued conversation about adding JPMS support.

saudet:
@HGuillemet Ah, here's one issue already. The generated sources as well as the ones in the `org.opencv` package are not getting included in the `sources` artifacts.

HGuillemet:
[https://github.com/bytedeco/javacpp-presets/pull/668](url)

HGuillemet:
Android Studio 3.3 may well have a fix for the problem, but it's currently available as a preview only.
Is it easy for you to test with Android Studio 3.3 ?
The only user-friendly way I see to support previous Android Studio versions and jpms is to use classifiers to provide a version of the jar without module-info.class.

HGuillemet:
[AS 3.3 list of changes](https://androidstudio.googleblog.com/2018/06/android-studio-33-canary-1-available.html)
It may be the jpms version that is provided as a classifier, rather than the other way around if you prefer. This is for javacpp. For javacpp-presets we can just exclude module-info from the android jar, or provide 2 android jars, with and without. And we maybe can adjust the dependencies in the pom towards javacpp to include the classifier or not depending on the platform.
.... or just request AS 3.3 for 1.5

saudet:
Ah, I see what you mean, it should be fixed:
> Processing module-info.class during desugaring analysis caused failures with the message java.lang.IllegalArgumentException: Type without superclass: module-info. This file is now filtered out during desugaring analysis.

That version has been available since June, so we should be OK...

Issue #16, opened 2015-05-23
GetPrimitiveArrayCritical option for byte array push


eix128:
Hello , there can be an option on JavaCPP.
That can pass parameter without copying.
Example like :

void doOperation(@Critical byte bigdata[]) {
    ... this byte array will be on C++ side without copying.
}

saudet:
That's for a use case where direct NIO buffers are out of the question?

saudet:
It's a good idea though, thanks for mentioning it! And it shouldn't be too hard to implement, so would you like to give it try?

jduke32:
what kind of help you want?

saudet:
For updating the code in `Generator`, and maybe writing a unit test to make sure it actually works.

jduke32:
did you look at ANTLR4 ?

saudet:
Yes, it still only supports context-free parsing, so it doesn't work well
with C++. What does that have to do with GetPrimitiveArrayCritical though?

jduke32:
it can be good for video processing.
passing images or audio without "copying"

jduke32:
memorymappedfile can be solution for that also ?

saudet:
Direct NIO buffers support all that, so why not use use those?

aishfenton:
A use-case for a GetPrimitiveArrayCritical over a direct buffer is that the direct buffer is very slow to access from the Java side (or at least, that's my understanding). So if you have code that needs to access the array at frequently from both the native and Java ends, then GetPrimitiveArrayCritical seems like the best bet (as long as you're ok with pausing the GC).

saudet:
@aishfenton That's not usually the case, no:
http://web.archive.org/web/20111117132310/http://blog.vlad1.com/2011/10/05/looking-at-java-nio-buffer-performance/

vincent-grosbois:
Hi
as discussed in #253 I'll try to implement it, but what do you think would be best?
I can  add a new  flag in the methodInfo instance or somewhere else (ie a global config) ?

saudet:
It can be both, like `@NoException`.

vincent-grosbois:
where in the code do you see that @NoException can be setup for all the methods?

saudet:
https://github.com/bytedeco/javacpp/blob/master/src/main/java/org/bytedeco/javacpp/annotation/NoException.java

vincent-grosbois:
I'm not sure I follow...
it seems the annotation can be added for types and method
[...]

PR #254, opened 2018-08-23
#16 created ForceArrayCritical annotation to prevent copies of array

vincent-grosbois:
* When this annotation is added to a method or a class, then all
wrapper calls will use Get/ReleasePrimitiveArrayCritical pair instead
of Get/Release<primitivetype>ArrayElements pair.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 
+ * Add new `@CriticalRegion` annotation to allow zero-copy access to data of Java arrays ([pull #254](https://github.com/bytedeco/javacpp/pull/254))
  * Allow `Builder` to create links for resource libraries even when no Java classes are built
  * Fix `Loader.cacheResource()` creating a subdirectory named "null" when caching a top-level file
  * Update `README.md` with reference to newly published [Mapping Recipes for C/C++ Libraries](https://github.com/bytedeco/javacpp/wiki/Mapping-Recipes)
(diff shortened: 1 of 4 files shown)


saudet:
Thanks! Let's also create a small test to make sure it works.

saudet:
Also let's make it behave like `@NoException`, unless you see a good reason not to?

saudet:
Hum, there's no test for `@NoException`, I'm just relying on the presets build to test that at the moment, so let's not worry about a test for the critical one either.

Anyway, as per the JDK's documentation, this function isn't about arrays, it basically stops the JVM:
https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/functions.html#GetPrimitiveArrayCritical_ReleasePrimitiveArrayCritical
Since this is really about the method itself and not about each argument a name like `@CriticalRegion` would reflect that better. It will also allow us to use transparently other available JNI functions such as `GetStringCritical()` and potentially others as they show up for other kinds of objects or for other purposes that also require a stopped JVM.

For this reason, it basically has the same semantics as `@NoException`, but instead of dealing with C++ exceptions, it enables a critical region in the JVM.

vincent-grosbois:
Hi @saudet 
I've updated my commit, hopefully it updates the PR as well.
I made it behave like NoException annotation, so that it will recursively look for the tag in parent classes.
Also renamed the annotation.

PR #256, opened 2018-09-06
Update android-arm.properties

zkrige:
Update linkpath. this should fix https://github.com/bytedeco/javacpp-presets/issues/38


saudet:
Thanks! I think we should also change that path in `platform.includepath` as well?

saudet:
Did you mean that it fixes https://github.com/bytedeco/javacpp-presets/issues/562? The other one is related to a loading issue, not a build issue.

zkrige:
weird - i copied the link exactly - not sure why github changed it

yes, will make sense to change the includepath as well. will do now

Issue #269, opened 2018-10-12
Some command lines in the README are not working


stonexjr:
For example:
```bash
$ java  -cp javacpp.jar NativeLibrary
```
simply throws error
`Error: Could not find or load main class NativeLibrary`
Need to add current directory to the classpath like:
```bash
$ java  -cp javacpp.jar:. NativeLibrary
```

saudet:
No, it works just fine because "." is in the class path already, see #192. In any case, we can't put `java  -cp javacpp.jar:. NativeLibrary` either because it doesn't work on Windows, we need to use a semi-colon on that platform. Anyway, I'll mark this as a duplicate of #192. Let's continue the discussion there! Thanks

PR #270, opened 2018-10-12
Fixed broken example command lines in the README file

stonexjr:
As detailed in issue #269


saudet:
BTW, which version of the JDK is causing this issue?

stonexjr:
I was using JDK 1.8

saudet:
Thanks! I've confirmed multiple times that it works fine with JDK 1.8 on Linux, Mac, and Windows, so there's something that's not typical about your installation. Most people will have a standard installation of the JDK, so that's not going to be problem. In any case, we need to do something about this some other way (https://github.com/bytedeco/javacpp/issues/269#issuecomment-429496299).

PR #273, opened 2018-12-03
Indexer rows(), col().... returning -1 if not 3D

HGuillemet:


saudet:
Looks good thanks! Could you change the target branch to `master` though? I'm just keeping the `jpms` branch here for any breaking changes related to JPMS, but otherwise will keep it up-to-date with `master`.

PR #275, opened 2018-12-04
rows(), cols()... now returning -1 if not 3D

HGuillemet:


saudet:
BTW, a semi-related question, I don't think this is actually true, what do you think?
> ```java
>      * {@code strides[i] > strides[i + 1] && strides[strides.length - 1] == 1} must hold.
> ```
https://github.com/bytedeco/javacpp/blob/master/src/main/java/org/bytedeco/javacpp/indexer/Indexer.java#L61

HGuillemet:
From the OpenCV doc about Mat:
> Note that M.step[i] >= M.step[i+1] (in fact, M.step[i] >= M.step[i+1]*M.size[i+1] ). This means that 2-dimensional matrices are stored row-by-row, 3-dimensional matrices are stored plane-by-plane, and so on. M.step[M.dims-1] is minimal and always equal to the element size M.elemSize() .

So for Indexer (when indexing a Mat), your assertion should hold.

saudet:
I know, but I don't think Indexer needs it...

PR #278, opened 2019-01-07
Use regex to translate Doxygen commands

HGuillemet:
Small suggestion to use regex for parsing Doxygen commands. 
Main use is to remove`[in]`, `[out]` or `[in,out]` from `@param`, but may be a starting point for a more complex parsing.

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/tools/Parser.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Parser.java
@@ -22,23 +22,17 @@
 
 package org.bytedeco.javacpp.tools;
 
+import org.bytedeco.javacpp.ClassProperties;
+import org.bytedeco.javacpp.Loader;
+
 import java.io.File;
 import java.io.FileNotFoundException;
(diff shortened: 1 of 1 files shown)


saudet:
We shouldn't remove any information though. When something doesn't show up
in Javadoc, we should also use some other format that it accepts so it gets
displayed, but this is secondary. Even when Javadoc complains the
information still shows up in the source code.

HGuillemet:
If you'd like to keep the direction parameter, we could translate `\param[in] xxx yyy`  to something like `@param xxx [in] yyy`.

HGuillemet:
Here it is. This PR also translates `\sa`. Only `\see` were translated. 
However obtaining a correct translation of all Doxygen grammar is quite impossible task. For instance some comments are lost because not at the right place in the .java. (`/** */` blocks before a C++ namespace do not appear in javadoc because not moved before `class`, also Doxygen allows comments concerning the method after `\param`, but javadoc will treat this comment as part of the last `@param`).

PR #281, opened 2019-01-15
Jpms fix empty param

HGuillemet:
Fix Doxygen \param translation when param description is empty

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 
+### January 11, 2019 version 1.4.4
  * Allow users to override platform properties via system properties starting with "org.bytedeco.javacpp.platform."
  * Have `BuildMojo` output its class path back to the Maven project as `${javacpp.platform.artifacts}`
  * Fix potential `NullPointerException` in `Loader.findResources()` under the bootstrap class loader
--- a/README.md
+++ b/README.md
@@ -154,9 +154,7 @@ public class NativeLibrary {
 
 After compiling the Java source code in the usual way, we also need to build using JavaCPP before executing it as follows:
 ```bash
-$ javac -cp javacpp.jar NativeLibrary.java 
-$ java -jar javacpp.jar NativeLibrary
-$ java  -cp javacpp.jar NativeLibrary
+$ java -jar javacpp.jar NativeLibrary.java -exec
 Hello World!
 ```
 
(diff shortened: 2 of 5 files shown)


saudet:
Thanks, looking better, but there are still a few issues with it:
- It converts things like `\\throw` into `\@throws`, but `\\throw` with 2 backslashes isn't a tag to start with
- Javadoc doesn't recognize `@param[in]`, `@param[out]`, `@deprecated.`, `@deprecated:`, and probably others as tags, so we should make sure to always include a space after the tag itself on output
- Javadoc doesn't display the information of `@see` tags when placed, for example, inside a `@param`

HGuillemet:
>     * It converts things like `\\throw` into `\@throws`, but `\\throw` with 2 backslashes isn't a tag to start with

Should be fixed in latest commit to this PR.

> 
>     * Javadoc doesn't recognize `@param[in]`, `@param[out]`, `@deprecated.`, `@deprecated:`, and probably others as tags, so we should make sure to always include a space after the tag itself on output

@param[in] case is covered by previous changes. For others, I added a check like it exists before using regex.
> 
>     * Javadoc doesn't display the information of `@see` tags when placed, for example, inside a `@param`

I think we can accept this. What would you suggest ?

saudet:
Thanks for the fixes, it's not converting `\\throw` into `\@throws` anymore, but now it's converting all `\\`  sometimes into `&#064;` (why?) and sometimes `\`, which I guess is intentional because Javadoc doesn't need to escape backslashes, is this the rationale? I'm pretty sure it breaks LaTeX equations though as `\\` usually starts a newline...

As for `@see` tags inside `@param` one, that's acceptable I think, yes. It doesn't occur that often, and the information is still in the source file anyway.

saudet:
Ah, I see, `\@` should get translated to `&#064;`, yes that makes sense. As for `\\` we should probably not translate to `\` unless we can know we're not in a LaTeX context.

HGuillemet:
Yes, it converts \@ to the HTML entity for @. We could use {@literal @} but it won't work in all context.

The LaTeX code will not be interpreted after translation to javadoc, will it ? But why not, we could keep the \\, it's a simple fix. Commit comming.

HGuillemet:
In fact it's the \f command that should be interpreted as verbatim.

saudet:
There's also at least $f ...

saudet:
Latest changes seem to work pretty well, thanks! The only thing that doesn't seem to be working is when the first `\f$` and second `\f$` tags are on different lines.

PR #305, opened 2019-05-20
Fix cacheResource to work with jlink images

HGuillemet:
When extracting resources from a jlink image, the `URLConnection` to the resource is an instance of `sun.net.protocol.jrt.JavaRuntimeURLConnection`. The loader failed to get a proper last modified date and content length in this case. This PR fixes this.

When reading from a `JavaRuntimeURLConnection`, `URLConnection.getContentLength` works, but` getLastModified` or `getDate` do not. We access the jrt file system and gets the last modified date of its root directory. The result is cached. I haven't found a simpler way to get the build time of the jlink image.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 
+ * Fix `Loader.cacheResource()` with the "jrt" protocol as used by jlink ([pull #305](https://github.com/bytedeco/javacpp/pull/305))
  * Fix compiler error with `SharedPtrAdapter` and `UniquePtrAdapter` in callback functions ([pull #304](https://github.com/bytedeco/javacpp/pull/304))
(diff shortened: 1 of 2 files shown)


saudet:
What happens if we call something like `Paths.get(new URI(resourceURL.toString())`?

HGuillemet:
After removing the fragmen  from the URL, we get a path like:
`/org.bytedeco.openblas.linux.x86_64/org/bytedeco/openblas/linux-x86_64/libopenblas.so.0`
If your idea is to get a path to a physical file, that won't do it.

saudet:
No, but Files.size() and getLastModifiedTime() probably work on it, right?

HGuillemet:
Ok, I see.
It does work indeed, baring [this bug](https://bugs.openjdk.java.net/browse/JDK-8216553).
It seems cleaner than my `getRuntimeBuildTime` method. I'll update the PR.

PR #306, opened 2019-05-20
Replace Path.equals check by Files.isSameFile

HGuillemet:
`Paths.equals` returns false for paths like `bin/../lib` and `lib`.
Replaced by Files.isSameFile when checking if a symlink must be created.
Prevents libs in cache to be overwritten with symlinks pointing to themselves.


saudet:
`Files.isSameFile()` doesn't work for non-existent files, so that prevents symbolic links from getting created at all. Maybe what we need is `Path.normalize()`. Can you try that?

HGuillemet:
Oops, you're right. Deleting this PR, another one to come.

PR #307, opened 2019-05-21
Add Path.normalize before symlink creation

HGuillemet:
Normalize path like `bin/../lib` to `lib` before checking for symlink creation.
Prevents libs in cache to be overwritten with symlinks pointing to themselves.
I have a doubt about lines 1482 and 1500:
```
 !relativeTarget.equals(linkPath2.getFileName())
```
Shouldn't that be:
```
 !relativeTarget.equals(linkPath2)
```
?
(same for line 1483).


saudet:
No, those are not absolute paths, so they can't be compared.

HGuillemet:
Ok, so removing the normalize() for those, they are useless.

PR #332, opened 2019-08-19
OSGi enabled JavaCPP

timothyjward:
Following on from discussions in PR #328 this PR *does not change* the Maven Plugin layout and

1. Adds OSGi metadata to the JavaCPP library
2. Adds an OSGi integration test

Note that the integration test passes because the JavaCPP `Loader.load()` method is not used, if it is used then the test fails. This demonstrates how the native library loading done by JavaCPP fails in OSGi, and hopefully can be used to design a fix.


pom.xml
```diff
@@ -157,19 +167,31 @@
...
+              <name>org/bytedeco/javacpp/</name>
+              <manifestEntries>
+                <Implementation-Title>${project.name}</Implementation-Title>
+                <Implementation-Vendor>Bytedeco</Implementation-Vendor>
+                <Implementation-Version>${project.version}</Implementation-Version>
+                <Specification-Title>${project.name}</Specification-Title>
+                <Specification-Vendor>Bytedeco</Specification-Vendor>
+                <Specification-Version>${project.version}</Specification-Version>
+              </manifestEntries>
+            </manifestSection>
```
saudet: It looks like because of this change this is not able to pick up the version number anymore:
```java
String version = Builder.class.getPackage().getImplementationVersion();
```
https://github.com/bytedeco/javacpp/blob/master/src/main/java/org/bytedeco/javacpp/tools/Builder.java#L1214
Any ideas why that might be?

pom.xml
```diff
@@ -278,6 +300,80 @@
...
+            <projectsDirectory>src/it/</projectsDirectory>
+            <streamLogs>true</streamLogs>
+            <localRepositoryPath>${project.build.directory}/integration-test/repo</localRepositoryPath>
+            <properties>
+              <maven.compiler.source>${maven.compiler.source}</maven.compiler.source>
+              <maven.compiler.target>${maven.compiler.target}</maven.compiler.target>
+              <org.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener>warn</org.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener>
+            </properties>
+            <goals>
+              <goal>verify</goal>
```
saudet: The "verify" phase isn't getting executing by default on "install". Is this expected?

saudet:
If I understand correctly OSGi wants to manage native libraries on its own with `Bundle-NativeCode`? It looks rather limited in functionality. We probably won't be able to use that. We'll need to figure out the right way to call `System.load()`.

pom.xml
timothyjward: So my intention with this was to maintain the correct the layout in the manifest. The existing intstructions include a `Name` header, which is a reserved header designed to tie the next header(s) to a specific package. Unfortunately when merging these values into an existing manifest the maven-jar-plugin was inserting the entries at random points and therefore resulting in a broken Manifest. Apparently I got far enough to make the manifest valid, but not quite all the way to making it pick up the package-specific values.

pom.xml
timothyjward: So the Invoker Plugin binds automatically to:
* the `pre-integration-test` phase for the `install` goal (https://maven.apache.org/plugins/maven-invoker-plugin/install-mojo.html)
* the `integration-test` phase for the `run` goal (https://maven.apache.org/plugins/maven-invoker-plugin/run-mojo.html)

This means that the integration tests will run as part of any top-level command which triggers the `integration-test` lifecycle phase (this includes both `verify` and `install`)

The `verify` phase that you've commented on is the lifecycle target of the sub-build. This can be seen in the nested output, but it may not be obvious. 

I do see the tests run when running when I issue `mvn clean install` at the top level

timothyjward:
> If I understand correctly OSGi wants to manage native libraries on its own with Bundle-NativeCode? It looks rather limited in functionality. We probably won't be able to use that. We'll need to figure out the right way to call System.load()

So what OSGi does with `Bundle-NativeCode` is to allow you to package multiple native binaries for different platforms and have the correct one selected for you automatically. Furthermore it sets up the bundle's class loader properly so that the native library can be found and loaded by the bundle class loader with a call to `System.loadLibrary(libName)` without any further file extraction required. 

I'm currently investigating the feasibility of generating a small Java Helper class to live alongside the class(es) using JavaCPP. This helper can be called to try to load the native library within the correct Class context. If you have another idea about how this could be achieved then I'd be happy to hear it!

pom.xml
timothyjward: > Any ideas why that might be?

I've just figured this out - the `Name` header is limiting the following:
```
Specification-Title: JavaCPP
Specification-Vendor: Bytedeco
Implementation-Vendor: Bytedeco
Specification-Version: 1.5.2-SNAPSHOT
Implementation-Version: 1.5.2-SNAPSHOT
Implementation-Title: JavaCPP
```
 to only apply to the `org.bytedeco.javacpp` package, which I assume was the original intent of having the named section. This PR has now "fixed" this to work properly, rather than getting inconsistent behaviour from the maven-jar-plugin. 

The bug comes in because the Builder is attempting to get the version of its own package, which is `org.bytedeco.javacpp.tools` not `org.bytedeco.javacpp`. This package has no section in the manifest and so is unversioned.

There are three possible fixes I can think of:
 1. Change the named section to point at the tools package
 2. Add a new named section for the tools package
 2. Remove the name from the section and have the manifest apply to all packages

Any preferences?

saudet: I'd rather just leave it like it was before (all packages). It doesn't matter if the entries in the manifest file don't end up in any meaningful order. They never do anyway and we don't often look at it manually either.

saudet:
Let's see, `System.load()` looks like this:
```java
    public static void load(String filename) {
        Runtime.getRuntime().load0(getCallerClass(), filename);
    }
```
https://github.com/openjdk-mirror/jdk7u-jdk/blob/master/src/share/classes/java/lang/System.java#L1058-L1060

Still seems to be the same on the tip:
https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/lang/System.java#L1871

Would you be against trying to call `Runtime.load0()`? Or is this pretty much guaranteed to fail with OSGi anyway?

pom.xml
saudet: Strange, doesn't do that here. Let me look at that later.

timothyjward:
> Would you be against trying to call Runtime.load0()? Or is this pretty much guaranteed to fail with OSGi anyway?

Obviously there are issues with reflectively calling the internals of the JDK (for example this may not work on Eclipse's J9) but there isn't anything that OSGi will do to stop you and it should work. My biggest concern is that this sort of thing may well not work in future versions of Java as the VM is further locked down. When it comes down to it it's your call. I'll create a PR for the "helper class" approach so that you can take a look.

saudet:
OpenJ9 uses OpenJDK, Android uses OpenJDK, there isn't much of anything that doesn't use OpenJDK these days, well maybe except IBM JDK, but they haven't updated since Java SE 8, so that's not a problem IMO. If OpenJDK ever decides to change that API, or even better, make the equivalent public, then we can always update at that point. There is still nothing to replace `Unsafe` completely, so people are still using, and the world hasn't ended. Just saying. The alternative is going to be much uglier.

saudet:
Looks good, but as I pointed out, we don't need to use try to use `Runtime.load0()`, unless the class loader are not the same. Let's add a check for that?

And could you revert the breaking changes to the manifest as well?

timothyjward:
> Looks good, but as I pointed out, we don't need to use try to use Runtime.load0(), unless the class loader are not the same. Let's add a check for that?

Done

> And could you revert the breaking changes to the manifest as well?

Also done, but not as a revert (which results in a broken OSGi bundle and fails the integration test). Instead I have applied the same fix as is on the `osgi` branch which removes the manifest section and applies the metadata to all the packages.

timothyjward:
It looks like the Travis build is passing. Are you happy to merge this now?

PR #331, opened 2019-08-19
Enable JavaCPP in OSGi

timothyjward:
Following on from discussions in PR #328 this PR takes the separated Maven Plugin layout and

1. Adds OSGi metadata to the JavaCPP library
2. Adds an OSGi integration test

Note that the integration test passes because the JavaCPP `Loader.load()` method is not used, if it is used then the test fails. This demonstrates how the native library loading done by JavaCPP fails in OSGi, and hopefully can be used to design a fix.

PR #335, opened 2019-08-22
Packaging fixes for the OSGi branch

timothyjward:
This PR:

* Further corrects the Manifest as discussed in #332 
* Fixes the Travis build issue identified in #331 so that the tests can run and snapshots be deployed

These changes are targeting the `osgi` branch, but they can easily be appended to #331 if there is a desire to include them in `master`.

### DIFF
--- a/javacpp-tests-osgi/pom.xml
+++ b/javacpp-tests-osgi/pom.xml
@@ -69,17 +69,6 @@
         <version>${project.version}</version>
         <executions>
           <execution>
-            <id>javacpp.parser</id>
-            <phase>generate-sources</phase>
-            <goals>
-              <goal>build</goal>
-            </goals>
-            <configuration>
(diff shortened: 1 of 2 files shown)


saudet:
The parser is not working in tests? That doesn't sound like something we should ignore.

timothyjward:
> The parser is not working in tests?

No, this was my misconfiguration of the JavaCPP Maven Plugin based on a misunderstanding of the documentation. I had both the `parse` and `build` goals configured, but I had only set the include path on the configuration for `build`. 

I'm not sure why I didn't see failures locally with this, but Travis was there to keep me honest. The simplest fix was to remove the `parse` goal from the project because it wasn't doing anything particularly useful (I have JavaCPP annotations on my class, so `build` is what I really want). To the best of my knowledge there are no actual issues with the `parse` goal.

PR #336, opened 2019-08-23
Fix Native Library Loading in Osgi with a Java Helper Class

timothyjward:
This PR is an option for solving the JavaCPP native library loading issue in a multi-classloader environment such as OSGi. As outlined in #332 the `Builder` has been extended to generate a small helper class which lives in the same package as the annotated class calling the `Loader` this can be used by the loader to ensure that the native library is loaded by the correct classloader.

A small number of API additions are required to ensure that the class context is available when loading the native library in the `Loader`, further additions are made to the `Builder` to allow configuration of the Java class generation.


saudet:
Yeah, that's ugly alright...

timothyjward:
> Yeah, that's ugly alright...

Agreed, but it avoids using the parts of Java we're not supposed to touch. It may therefore be useful in future if the VM starts to forbid `setAccessible(true)`. I mostly wanted to prove that this solution was possible, and that it fixed the issue, so that you would have some alternatives available.

PR #337, opened 2019-08-23
Fix native library loading in OSGi using Runtime load0

timothyjward:
As discussed in #332, this is an alternative implementation for establishing the correct class context when loading native code. See also #336.

Note that the same API additions need to be made to `Loader` so that Class context can be passed in.

Signed-off-by: Tim Ward <timothyjward@apache.org>

### DIFF
--- a/javacpp-tests-osgi/src/main/java/org/bytedeco/javacpp/test/osgi/Calc.java
+++ b/javacpp-tests-osgi/src/main/java/org/bytedeco/javacpp/test/osgi/Calc.java
@@ -12,17 +12,17 @@ public class Calc {
         // isn't because the library gets loaded by
         // the wrong classloader
         //
-        // Loader.load();
+        Loader.load();
         
         // This is what we need to happen in the scope
         // of the bundle containing the native code, 
         // not the JavaCPP bundle. Note that the call
         // to the Loader is just to force a wiring to
         // the JavaCPP package needed by the JNI code
-        
-        Loader.getPlatform();
-        
-        System.loadLibrary("jniCalc");
+        //
+        //Loader.getPlatform();
+        //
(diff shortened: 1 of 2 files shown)


saudet:
Not too ugly, but we get `WARNING: An illegal reflective access operation has occurred` on JDK 9+...

timothyjward:
> Not too ugly, but we get WARNING: An illegal reflective access operation has occurred on JDK 9+...

Agreed - this is something that is supposed to get killed off in future versions of Java, (i.e. `setAccessible(true)` will throw an Exception)

saudet:
How about this: We check if the Classloader of Loader and the class we're trying to load is the same, and if it's the same, we keep doing like before, but if it's different, we use that hack. This way, if/when it starts affecting enough people, maybe OpenJDK will become more responsive in giving us something usable, but in the meantime it won't affect existing users.

timothyjward:
That sounds fine to me. I won’t be able to work on this for the next week, so if you’d like to make the change and merge then please go ahead, otherwise I’ll update the PR at the start of September.

saudet:
No problem! How about we merge pull #336 to the `osgi` branch and integrate this approach to pull #332 for the `master` branch?

timothyjward:
I've taken a version of the commit in this PR (including the class loader check) and pushed it to #332

timothyjward:
> How about we merge pull #336 to the `osgi` branch

I'm happy for that to be the case, but it seemed as though you felt Runtime.load0 was the preferred approach?

saudet:
Right, but I'm thinking we can keep that around in the `osgi` branch with the rest of the breaking changes that we might need to consider as some point. This way people can try it out and see if this is really something that we need.

PR #339, opened 2019-09-01
Add thread name when AttachCurrentThread

louxiu:
It would be better to use java thread name that matched native thread id for debugging.

PS: Because I don't have android env, so I only test setting thread name on mac.


saudet:
Your pull request only sets the name in the case of Android though. Are there some commits missing?

louxiu:
@saudet Patch append.

Issue #342, opened 2019-09-19
default values for function arguments need template substitution


mkmckenz:
If the default value of a function argument uses a template parameter, it is not resolved in the generated JNI .cpp which then fails to compile.  For example:

```cpp
template <class T, class CompareFunctorT>
struct SortedContainer {
    SortedContainer(CompareFunctorT compareFun = CompareFunctorT());
};
```
And lets say I bind an instantiation of that template like so:
```java
 map.put(new Info("SortedContainer<std::string, std::less<std::string> >").pointerTypes("SortedStringContainer"));
```
Because Parser.parameters() does not substitute template arguments for template parameters used in default values, the generated JNI C++ code for the bound instantiation of the template above will contain the expression `ptr0 == NULL ? std::less<std::string>(CompareFunctorT()) : *ptr0`, which doesn't compile and should instead be `ptr0 == NULL ? std::less<std::string>() : *ptr0`.

saudet:
Fix included in JavaCPP 1.5.2! Thanks for the contribution.

Issue #51, opened 2015-12-03
Improve Parser: Use the Clang API


Arcnor:

Arcnor:
I'm in the process of doing this right now. Currently, the following issues exists with the approach I'm taking:
* Preprocessor directives are not supported: This is fixable, because Clang can be made aware of those, but it doesn't by default. I'll leave this for later.
* Normal comments cannot be read: Unless I've missed some Clang flag, non-doc comments are completely ignored. I'm not sure this is a big deal, but nonetheless, it's different from the current behavior.

saudet:
BTW, we probably want to use the C++ API of Clang for this. It is not currently mapped by the presets, so as initial work, we would either have to:

1. Code the new parser temporarily in C++, or
2. Create the presets for the C++ API of Clang, using the current `Parser`.

Either way is fine with me. Thanks for your interest in this project and let me know how I can help!

Arcnor:
Sorry, I haven't had the time to work on this lately.

My final changes allowed me to parse a lot, but some missing things from the C++ API prevented me finishing it IIRC, so I think option 2 should help us get there (or, as you said, code it in C++, but it's non-trivial :D)

Arcnor:
I've continued a bit on this, and for now, I've decided to write the Clang bindings manually, as I think the surface we need from Clang is not that big (I might be wrong, though).

Once we hace a working parser, we can generate proper bindings for Clang itself and use them in the generator, closing the circle ;)

I have some doubts about how to implement some of the bindings, though, so I'll hit the forums in a few days with my questions :)

saudet:
@Arcnor Any progress with this?

libclang seems to be getting pretty good for that sort of thing, for example:
https://github.com/rust-lang-nursery/rust-bindgen
https://rust-lang-nursery.github.io/rust-bindgen/
So maybe we won't need to use the C++ API after all...

Arcnor:
Hi Samuel,

No, unfortunately I haven't had the time to continue, not enough incentive for me to do so right now (the project(s) that were using JavaCPP all stopped for one reason or another, mostly priorities).

So `libclang` is getting better, eh? That's great news! I've had a quick look at the API again, and it seems to contain some goodies I don't remember from 2 years ago, so yeah, maybe now it's enough for our purposes.

[...]

PR #343, opened 2019-09-19
Perform template substitution when parsing default values

mkmckenz:
For reference, here's how I've patched for issue #342.  This change worked for me, but I won't claim it's the **correct** fix.


saudet:
Looks good! Thanks. The Parser is pretty ad hoc at the moment, so whatever works is probably "correct". :) Long term, the idea is to use something like Clang, see issue #51.

PR #344, opened 2019-09-25
Add further OSGi testing to validate JavaCPP functionality

timothyjward:
* Initial fix to ensure class context when putting offsets
* Remaining bug where JavaCPP Pointer types cannot access native functions

Signed-off-by: Tim Ward <timothyjward@apache.org>

This Pull Request came about as I started to use JavaCPP in more advanced use cases. I found that the calls to `Loader#putMemberOffset` were failing because they were attempting to use JavaCPP's classloader to load all referenced types. This works fine for the built in pointer types, but not for Custom Pointer types defined for the application.

I fixed this by establishing a link between the context class and the thread loading the library, which avoids these problems, however I then immediately hit another issue. Because the generated `jnijavacpp` code is being loaded by a different classloader types such as `IntPointer` are unable to access their own native methods. Fixing this will be somewhat more challenging, and I would appreciate some input from JavaCPP experts!

I'm happy to port this test over to the master branch if that would be helpful, but for now I've worked on top of the OSGi branch as it's somewhat easier to develop/test.


saudet:
I think the general idea makes sense. Let's use `Deque` for the stack instead of `Stack` though. Could you also use a simpler test instead of copy/pasting one of the existing ones? Or reuse it somehow without having 2 copies checked in?

As for `jnijavacpp`, the idea eventually is to make it available as a separate library, as part of JavaCPP itself, using a build process similar to the presets. This would also fix corner cases where users try to use one of the classes from JavaCPP before loading anything else.

timothyjward:
> I think the general idea makes sense. Let's use Deque for the stack instead of Stack though.

Done

> Could you also use a simpler test instead of copy/pasting one of the existing ones? Or reuse it somehow without having 2 copies checked in?

Reusing the test sources would have needed some moderately hairy maven config, so I've just cut down the test a lot (focussed on my original class loading issue). The tests still fail due to the `jnijavacpp` lib not being loaded in the JavaCPP classloader. 

I'm not sure of the best approach to fix this additional bug though. As you've stated it would need `jnijavacpp` to be available as a separate library, and for it to be loaded early on somehow in the JavaCPP classloader. I'm not certain, but it would probably *also* need to be loaded in the class loader of the target class as I think there are functions from it that are called in the other JNI libraries generated by JavaCPP.

Any thoughts as to how that could be achieved?

saudet:
> > I think the general idea makes sense. Let's use Deque for the stack instead of Stack though.
> 
> Done
> 
> > Could you also use a simpler test instead of copy/pasting one of the existing ones? Or reuse it somehow without having 2 copies checked in?
> 
> Reusing the test sources would have needed some moderately hairy maven config, so I've just cut down the test a lot (focussed on my original class loading issue). The tests still fail due to the `jnijavacpp` lib not being loaded in the JavaCPP classloader.

The files are still mostly copy/paste :) If we just want a smoke test for the loader, I'm sure we can come up with a simpler one than that ... Anyway, let's do that when we get to merge something in the `master` branch. It doesn't really matter for the `osgi` branch as when we decide to break backward compatibility, I'll probably have better ideas about how to name the new modules and stuff.

> I'm not sure of the best approach to fix this additional bug though. As you've stated it would need `jnijavacpp` to be available as a separate library, and for it to be loaded early on somehow in the JavaCPP classloader. I'm not certain, but it would probably _also_ need to be loaded in the class loader of the target class as I think there are functions from it that are called in the other JNI libraries generated by JavaCPP.
> 
> Any thoughts as to how that could be achieved?

Actually, we can't load the same JNI library in 2 different class loaders. I'm not sure what you're referring to here though. `jnijavacpp.cpp` would still get linked into all JNI libraries created by JavaCPP, we'd just get an additional shared library that could be loaded independently in another class loader.

timothyjward:
> Actually, we can't load the same JNI library in 2 different class loaders. I'm not sure what you're referring to here though. jnijavacpp.cpp would still get linked into all JNI libraries created by JavaCPP, we'd just get an additional shared library that could be loaded independently in another class loader.

This would be great - I don't know how hard it is to make the builder output the extra library file though.

saudet:
We'd have to port stuff from the presets, probably messy, but still hoping
for something cleaner...

saudet:
Maybe this is where a couple of Maven Archetypes could start to become handy?

saudet:
So, do you really want to merge this into the osgi branch? I was under the impression that we wouldn't be maintaining that branch until we figure out something better and instead test things out without breaking backward compatibility in the master branch. To maintain the osgi branch, you'd also need apply all the other bug fixes there as well...

timothyjward:
> Maybe this is where a couple of Maven Archetypes could start to become handy?

That does sound like a good idea. Having a parent pom to set up the profiles properly and an archetype to make getting started easy would be a nice experience.

> So, do you really want to merge this into the osgi branch?

These changes could be adapted (i.e. pointed at the correct files) and added to master. This would fix the existing class loading bug (the one fixed with the `Deque`) but the extra test would need to be ignored until the remaining issue is fixed, otherwise this would break the build in master.

I'm happy to do this, but I won't have time available for a week or two due to a deadline crunch.

saudet:
Ok, please send a pull request against master when you get the time! Thanks

Issue #312, opened 2019-05-30
Bindings using UniquePtrAdapter fail to compile with GCC 7.4


GeekOffTheStreet:
When using the std:: namespace for unique_ptr, bindings for unique_ptr return values fail to compile on Ubuntu Bionic Beaver using gcc 7.4.

Representative error is below:
```
jniConfig.cpp: In member function 'virtual std::unique_ptr<config::IConfig> JavaCPP__0003a_0003aconfig_0003a_0003aIConfigService::createConfig(const char*)':
java/build//jniConfig.cpp:1049:95: error: conversion from 'UniquePtrAdapter<config::IConfig>' to 'std::unique_ptr<config::IConfig>' is ambiguous
     return UniquePtrAdapter< config::IConfig >((config::IConfig*)rptr, rsize, rowner);
                                                                                               ^
/java/build//jniConfig.cpp:535:5: note: candidate: UniquePtrAdapter<T>::operator UniquePtrAdapter<T>::U&&() [with T = config::IConfig; UniquePtrAdapter<T>::U = std::unique_ptr<config::IConfig>]
     operator U&&() { return UNIQUE_PTR_NAMESPACE::move(uniquePtr); }
     ^~~~~~~~
/java/build//jniConfig.cpp:534:5: note: candidate: UniquePtrAdapter<T>::operator UniquePtrAdapter<T>::U&() [with T = config::IConfig; UniquePtrAdapter<T>::U = std::unique_ptr<config::IConfig>]
     operator U&() { return uniquePtr; }
     ^~~~~~~~
Exception in thread "main" java.lang.RuntimeException: Process exited with an error: 1
	at org.bytedeco.javacpp.tools.Builder.generateAndCompile(Builder.java:644)
	at org.bytedeco.javacpp.tools.Builder.build(Builder.java:1095)
	at org.bytedeco.javacpp.tools.Builder.main(Builder.java:1315)
```

Updating Generator.java to remove `operator U&() { return uniquePtr; }` resolves the compilation issue.

saudet:
That just means we need additional casting for this to work, for example:
`@Cast({"config::IConfig*", "std::unique_ptr<config::IConfig>"}) @UniquePtr`
seems to be what your function needs.

It's also possible to create a new annotation like `@MoveUniquePtr`:
https://github.com/bytedeco/javacpp-presets/blob/master/tensorflow/src/main/java/org/bytedeco/tensorflow/presets/tensorflow.java#L964

[...]

PR #353, opened 2019-10-23
Fix unique_ptr return by value

GeekOffTheStreet:
This is a change that addresses the issues we were having in issue #312.


saudet:
I don't think we can use that in general... Is it the second const or the first const that matters though? The second one looks harmless enough.

GeekOffTheStreet:
Looks like just making the method const is sufficient (updated).

Issue #349, opened 2019-10-06
Raspberry Pi failure


peardox:
Yes, I know you already know but that was tagged on to the HDF5 stuff so proper bug report time

Is there still a skip tests switch? I  wanna do presets hdf5 for you.

This is an ARMHF 32bit on a Raspberry Pi4 with stock OS (a Debian Buster fork)

Of note is this which occurs right at the start - possibly the problem? Full log attached.

[debug.jcpp.zip](https://github.com/bytedeco/javacpp/files/3694898/debug.jcpp.zip)

WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.google.inject.internal.cglib.core.$ReflectUtils$1 (file:/usr/share/maven/lib/guice.jar) to method java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain)
WARNING: Please consider reporting this to the maintainers of com.google.inject.internal.cglib.core.$ReflectUtils$1
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release

saudet:
Duplicate of #347

saudet:
That's already fixed in commit https://github.com/bytedeco/javacpp/commit/606ac566a6837bd971318834e88be343e8bd080f. Please make sure to get the latest code.

peardox:
I got the latest code, in fact that server saw daylight < 12 hours ago and, yes - I can see your fix...

It don't work

Presets is having issues as well on armhf

saudet:
@peardox Would it possible for you to find out on which line of `PointerTest.java` it crashes?

peardox:
The problem appears to lie with the Debian Buster / Pi4 Java (log attached above)

I've been experiencing the same startup issues when trying a Tensorflow build for 32bit (resorted to cross compilation - there's a first time for everything)

When doing a Bazel build from source the same junk happens

The ...
```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.google.inject.internal.cglib.core.$ReflectUtils$1 (file:/usr/share/maven/lib/guice.jar) to method

```Stuff only occurs on the stock OS

saudet:
I see that you're trying to use OpenJDK, but that's not well supported for ARM. Make sure to obtain Oracle JDK and use that instead.

peardox:
Does the version of JDK matter?

[...]

PR #355, opened 2019-11-09
Use new Travis ARM64 infrastructure to run javacpp tests

vb216:
Could help test various openjdk and arm/amd64/win/mac etc permutations. 
Just added arm for now as the others would be less trivial
Should help in relation to #349

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 dist: xenial
-sudo: false
+sudo: true
 cache:
   directories:
     - $HOME/.m2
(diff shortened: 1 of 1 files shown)


saudet:
Looks good, thanks! Could you do the same for https://github.com/bytedeco/javacv? (OpenJDK 8 on Travis CI doesn't seem to include JavaFX, and they're not doing Oracle JDK 8 anymore, so that's why it's set to OpenJDK 11.)

PR #390, opened 2020-04-13
Fix typo

matteodg:
Very small fix: `channels()` is supposed to return -1 when `sizes.length!=3`, but instead is returning 1.

Out of curiosity is there a reason why in case of `sizes.length==2` we do not have `rows()` and `cols()` returning respectively `size[0]` and `size[1]`?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 
+ * Deprecate but also fix `Indexer.rows()`, `cols()`, `width()`, `height()`, and `channels()` ([pull #390](https://github.com/bytedeco/javacpp/pull/390))
  * Fix `Parser` producing invalid wrappers for basic containers like `std::set<std::pair<...> >`
  * Add compiler options for C++98, C++03, C++14, and C++17 to platform properties files ([pull #389](https://github.com/bytedeco/javacpp/pull/389))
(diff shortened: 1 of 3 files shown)


saudet:
No reason in particular. I guess that's another bug introduced by @HGuillemet. Please fix it? :)

HGuillemet:
The 1 instead of -1 is a typo indeed. But at least it doesn't trigger a SIGSEGV like the previous value  `sizes[2]` written by @saudet could :P 

For the case `sizes.length == 2` I don't know. I don't think there is a logic we could stick to to find proper values for these shortcut methods. I'd vote for deprecating or removing them.
In the general n-dimensional case, the meaning of each dimension could be anything : with or without channel, NHWC/CHWN/NCHW...

matteodg:
> For the case `sizes.length == 2` I don't know. I don't think there is a logic we could stick to to find proper values for these shortcut methods. I'd vote for deprecating or removing them.
> In the general n-dimensional case, the meaning of each dimension could be anything : with or without channel, NHWC/CHWN/NCHW...

I agree, I vote too to deprecate them as they can be misleading: I was tripped up as well moving from a 2D to a 3D world of coordinates 😄

Issue #398, opened 2020-05-20
javacpp when compiles C++ files includes JDK paths which are not same as system Java


krzysg:
This is not a big problem but since I thought that I did something wrong (like wrong env settings etc.) it is worth to mention this issue for the others that might encounter same problem.

I have multiple java JDKs installed on my system (Ubuntu/Linux) and currently I am using Java 11:
```bash
$ echo $JAVA_HOME 
/usr/lib/jvm/jdk-11.0.7/

$ mvn -v
Apache Maven 3.6.3
Maven home: /usr/share/maven
Java version: 11.0.7, vendor: Oracle Corporation, runtime: /usr/lib/jvm/jdk-11.0.7
Default locale: en_US, platform encoding: UTF-8
OS name: "linux", version: "5.4.0-29-generic", arch: "amd64", family: "unix"

$ java -version
java version "11.0.7" 2020-04-14 LTS
Java(TM) SE Runtime Environment 18.9 (build 11.0.7+8-LTS)
Java HotSpot(TM) 64-Bit Server VM 18.9 (build 11.0.7+8-LTS, mixed mode)
```

I have noticed that paths used by javacpp when compiling c++ srouce files are set for different JDK (jdk-14.0.1): 
```java
[INFO] g++ -I/home/gonciarz/ws/repo/JavaCppTemplate/src/main/cpp/pl/krzysg/ -I/usr/lib/jvm/jdk-14.0.1/include -I/usr/lib/jvm/jdk-14.0.1/include/linux /home/gonciarz/ws/repo/JavaCppTemplate/target/classes/pl/krzysg/cppLib/target/jniMyCppLibConfig.cpp /home/gonciarz/ws/repo/JavaCppTemplate/target/classes/jnijavacpp.cpp -march=x86-64 -m64 -O3 -s -std=c++14 -Wl,-rpath,$ORIGIN/ -Wl,-z,noexecstack -Wl,-Bsymbolic -Wall -fPIC -shared -o libjniMyCppLibConfig.so
```

It seems that when searching ```jni.h``` nad ```jni_md.h``` default/system java paths might not be used if there are other java JDK installed. The search is performed by this method:

https://github.com/bytedeco/javacpp/blob/3189a117080ca63ad93c6427cd367ed7395d5103/src/main/java/org/bytedeco/javacpp/tools/Builder.java#L116

Maybe it would be good to choose ```jni.h``` and ```jni_md.h``` from system java if possible?  
[...]

PR #400, opened 2020-05-26
Fixed java home path for latest JDKs from Oracle

krzysg:
Fix of java home path reported in https://github.com/bytedeco/javacpp/issues/398.

Tested on Linux (Java 8, 11, 14) and MacOS (Java 7, 9, 13).
Unfortunately I have no suitable Windows machine.

One thing that could be removed during merge are lines (I left it as it is since I do not understand the reason of it):
```java
                if (new File(dir, jvmlink).exists()) {
                    jvmpath.add(dir.getAbsolutePath());
                }
```
I do not see any reason to search for ```jvmlink``` file since ```jvmlink``` is a (at least on MacOS and Linux) string with ```-ljvm``` which is option for linker and not file to be found.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 
+ * Fix `Builder` for different "java.home" path returned by latest JDKs from Oracle ([pull #400](https://github.com/bytedeco/javacpp/pull/400))
  * Refactor `Builder` a little to work around issues with Gradle
  * Log as warnings `SecurityException` thrown on `Loader.getCacheDir()` instead of swallowing them
  * Fix memory leak that occurs with "org.bytedeco.javacpp.nopointergc" ([issue bytedeco/javacpp-presets#878](https://github.com/bytedeco/javacpp-presets/issues/878))
--- a/src/main/java/org/bytedeco/javacpp/tools/Builder.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Builder.java
@@ -126,6 +126,7 @@ public class Builder {
                        "jvm" + properties.getProperty("platform.library.suffix", "");
         final HashSet<String> jnipath = new HashSet<>();
         final HashSet<String> jvmpath = new HashSet<>();
+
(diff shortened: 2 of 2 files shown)


saudet:
Thanks! We'll need to make this work on Mac with JDK 8 too though:
https://travis-ci.org/github/bytedeco/javacpp/jobs/691390877

Yes, we sometimes need a path for the link library as well. What's the question?

krzysg:
> Yes, we sometimes need a path for the link library as well. What's the question?

Maybe I wasn't clear enough. 

Besides searching for ```jni.h``` and ```jni_md.h``` there is also search for paths where files with platform dependent names defined in ```jvmlink``` and ```jvmlib``` can be found (at least this is what code is doing).  
These variables are defined as:
 
```java
final String jvmlink = properties.getProperty("platform.link.prefix", "") +
                       "jvm" + properties.getProperty("platform.link.suffix", "");
final String jvmlib  = properties.getProperty("platform.library.prefix", "") +
                       "jvm" + properties.getProperty("platform.library.suffix", "");
```

For example ```jvmlib``` has value  ```libjvm.dylib``` in MacOS or ```libjvm.so``` in Linux, so these are names of library files that can be found so in a result we have library path. 

However  ```jvmlink```  has value ```-ljvm``` and this is one of the compiler/linker options in command line and not file to be found anywhere, so my question is why we are searching for it?

saudet:
> However `jvmlink` has value `-ljvm` and this is one of the compiler/linker options in command line and not file to be found anywhere, so my question is why we are searching for it?

Ah, I see what you mean. That's for platforms like Windows where `jvmlink` ends up being "jvm.lib", which is a file, and `jvmlib`, "jvm.dll", which is another file that might not be in the same directory, but in that case we're only interested by "jvm.lib" one.

krzysg:
> That's for platforms like Windows where `jvmlink` ends up being "jvm.lib"

Thanks for info, now it makes sense.

Issue #156, opened 2017-01-02
Unsafe get, put and array access


sirinath:
Due to unsafe methods getting intensified it is faster than native. So perhaps it might be an idea to replace where appropriate with Unsafe use.

Also using Unsafe to access Java arrays is faster since there not no bound checking, memory fence and type checking.

Some native methods may be replaceable with pure Java with Unsafe.

saudet:
Yes, the indexer package could be enhanced in many ways. As usual, contributions are welcome!

PR #408, opened 2020-06-22
WIP: Use sun.misc.Unsafe in pointers and not in indexers

matteodg:
`sun.misc.Unsafe` is used in raw indexer, but it should be really used in Pointers: the indexers are just a layer on top of array/buffer/pointers.

Still unfinished as it is not clear how to handle bool, clong, size_t pointers.

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/BoolPointer.java
+++ b/src/main/java/org/bytedeco/javacpp/BoolPointer.java
@@ -71,7 +71,6 @@ public class BoolPointer extends Pointer {
     public BoolPointer() { }
     /** @see Pointer#Pointer(Pointer) */
     public BoolPointer(Pointer p) { super(p); }
-    private native void allocateArray(long size);
 
     /** @see Pointer#position(long) */
     @Override public BoolPointer position(long position) {
@@ -90,7 +89,7 @@ public class BoolPointer extends Pointer {
     public boolean get() { return get(0); }
     /** @return the i-th {@code bool} value of a native array
      * @param i*/
-    @Cast("bool") public native boolean get(long i);
+    @Cast("bool") public boolean get(long i) { return Raw.getInstance().getBool(address + i * sizeof()); }
     /** @return {@code put(0, b)} */
     public BoolPointer put(boolean b) { return put(0, b); }
     /**
@@ -100,5 +99,8 @@ public class BoolPointer extends Pointer {
      * @param b the {@code bool} value to copy
      * @return this
      */
-    public native BoolPointer put(long i, boolean b);
+    public BoolPointer put(long i, boolean b) {
+        Raw.getInstance().putBool(address + i * sizeof(), b);
+        return this;
+    }
 }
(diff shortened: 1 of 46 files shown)


saudet:
This is merely duplicating what we already have in Indexer. If this is all the functionality you need, please use Indexer instead.

matteodg:
I'm using Indexer as you suggested, but maybe I'm missing something: isn't it the Indexers just a layer on top of array/buffer/pointer? Why are they accessing internal pointer structures using Unsafe, shouldn't the Indexers just call pointer.get()/pointer.put()?

matteodg:
Another example: I'm using the HDF5 JavaCPP preset and it takes and returns pointers objects, why I should wrap the pointer in an Indexer to access its data? I would like just to consider the pointer as a basic data storage (just like array or buffers), not a Indexer with a shape and indexing.

saudet:
Yes, Indexer is basically a wrapper for arrays, buffers, and pointers, and we need to use `sun.misc.Unsafe` for arrays and buffers as well, it's not just a pointer thing...

matteodg:
I cannot find where Unsafe is used to access arrays or buffers: I'm looking at `FloatArrayIndexer`and `FloatBufferIndexer` for example, I guess I'm missing something obvious here

saudet:
> Another example: I'm using the HDF5 JavaCPP preset and it takes and returns pointers objects, why I should wrap the pointer in an Indexer to access its data? I would like just to consider the pointer as a basic data storage (just like array or buffers), not a Indexer with a shape and indexing.

That's what `Pointer` does, using JNI. How would you use `sun.misc.Unsafe` to access something like `std::map<std::string,std::string>`?

If you can do everything you need with `Indexer` though, what is the issue?

> I cannot find where Unsafe is used to access arrays or buffers: I'm looking at `FloatArrayIndexer`and `FloatBufferIndexer` for example, I guess I'm missing something obvious here

https://github.com/bytedeco/javacpp/blob/master/src/main/java/org/bytedeco/javacpp/indexer/ByteArrayIndexer.java and we would also need it for `Buffer` if/when we decide to add optimizations to skip bounds checking, see issue #156.

matteodg:
OK, I see: for primitive pointers (whose element is a primitive in Java) we can actually use Unsafe for get() and put(), but for more complicated pointers (for any struct for example `std::map<std::string,std::string>`) I guess you need JNI because they might be aligned differently in different architectures? Did I understand correctly this is your point?

saudet:
Right, the definition of classes like `std::map<std::string,std::string>` is implementation dependent, so we basically need to run C++ code to access them.

If you're just worried about creating an instance of `Indexer`, we could add those as static methods like `FloatIndexer.get(FloatPointer p, int i)`, which would be a lot easier to maintain and I think would also make more sense.

matteodg:
I was actually more concerned from an architecture point of view: I see now Pointers in general have been created for the JavaCPP bridge purposes, so to account for structs/classes/unions in C++.
But in hindsight I was considering primitive Pointers being like a drop-in replacement of Java Buffers, so I expected them to be  as fast as possible. What about creating a PrimitivePointer abstract class as a super class for all primitive pointers and have them use Unsafe for get()/put()?

And have Indexers just a layer on top of arrays/Buffers/Pointers/MemoryHandles/etc. for shaping and indexing purposes.
In fact Indexers are not really used to deal with structs/classes/unions from C++.

saudet:
I'm repeating myself, but we need `sun.misc.Unsafe` for buffers and arrays as well. What you're proposing makes it harder to maintain. You could fork JavaCPP and maintain that if you want, but someone will need to maintain it and that person is not me.

saudet:
Also, don't forget to consider what to do when `sun.misc.Unsafe` is not available.

PR #414, opened 2020-07-15
Add support for static libraries with RoboVM

ErikGro:
Linking static libraries with JavaCPP in RoboVM currently is not working.

The JNI Invocation API from currently developed [RoboVM project](https://github.com/MobiVM/robovm)
is based on [Java 7 implementation](https://docs.oracle.com/javase/7/docs/technotes/guides/jni/spec/invocation.html). 
Support for System.loadLibrary() for static libraries was just added in [Java 8](https://docs.oracle.com/javase/8/docs/technotes/guides/jni/spec/invocation.html). 

The current approach from JavaCPP for RoboVM is initializing JavaCPP via JNI_Onload_Libraryname which never gets executed since roboVM does not support it.
To solve this problem I came up with a simple wrapper function which eventually calls JNI_Onload.

Since the library gets statically linked, I was not sure which name to return in load().

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/Loader.java
+++ b/src/main/java/org/bytedeco/javacpp/Loader.java
@@ -1239,6 +1239,12 @@ public class Loader {
             }
         }
 
+        if (properties.getProperty("platform").startsWith("ios") 
+        && Boolean.parseBoolean(properties.getProperty("platform.library.static", "false"))) {
+            Loader.robovmOnLoad();
+            return "roboVMstaticLib";
+        }
+
         String executable = p.getProperty("platform.executable");
         if (executable != null && executable.length() > 0) {
             String platform = p.getProperty("platform");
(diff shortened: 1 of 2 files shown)


saudet:
We can do the same by setting `platform.library.static` to false. Have you tried that?
https://github.com/bytedeco/javacpp/blob/master/src/main/resources/org/bytedeco/javacpp/properties/ios-arm.properties#L34

ErikGro:
This should be easier.  But unfortunately when initializing the classes in OnLoad, env->FindClass returns null for some reason. Interestingly this only fails in OnLoad. When calling env->FindClass in custom native functions it finds the classes for Loader etc.. So i guess it's not a problem of the classpath. Do you have an idea what could be the cause?

saudet:
Sounds like an issue with RoboVM? Did you try to ask them about that?

ErikGro:
Updates from RoboVM:
I talked to one of the main developers and he fixed the issue and created a [PR](https://github.com/MobiVM/robovm/pull/507).
It was related to using the wrong ClassLoader in Jni_Onload.
I will keep you updated when this PR is merged.
Then i might open a new PR and edit the properties file for iOS to work with dynamic libraries instead of static libraries.

saudet:
@johanvos Does Gluon VM work with dynamic libraries as well these days?

saudet:
Since the PR for RoboVM was merged, I'm assuming everything is working fine now, both for static and dynamic libraries, but please let me know if it's still not the case. Thanks for looking into this!

Issue #446, opened 2020-12-30
load issues on Android since 1.5.4?


tmm1:
continuing discussion from https://github.com/bytedeco/javacpp/pull/413#issuecomment-752285429

seems to be more than just warnings for me, i'm not getting any javacpp functionality.

```
12-29 18:19:56.555 19941 19941 W System.err: Warning: Could not load Pointer: java.lang.UnsatisfiedLinkError: dalvik.system.PathClassLoader[DexPathList[[zip file "/system/framework/android.test.mock.jar", zip file "/system/framework/android.test.runner.jar", zip file "/mnt/expand/4bc1480a-802e-4602-9636-c9c878864962/app/app-xnWv8azJJ4Z1P1r3Wq3SZw==/base.apk"],nativeLibraryDirectories=[/mnt/expand/4bc1480a-802e-4602-9636-c9c878864962/app/app-xnWv8azJJ4Z1P1r3Wq3SZw==/lib/arm64, /mnt/expand/4bc1480a-802e-4602-9636-c9c878864962/app/app-xnWv8azJJ4Z1P1r3Wq3SZw==/base.apk!/lib/arm64-v8a, /system/lib64]]] couldn't find "libjnijavacpp.so"
```

Maybe the `BufferPoolMXBean` is unrelated and the problem is it is trying to look for `libjnijavacpp.so`? Is that file supposed to exist?

When I compile at the moment, I get a single `libjniHDHRLib.so` which includes `jnijavacpp.cpp`:

```
Info: Generating app/src/main/jniLibs/arm64-v8a/jnijavacpp.cpp
Info: Generating app/src/main/jniLibs/arm64-v8a/jniHDHRLib.cpp
Info: Compiling app/src/main/jniLibs/arm64-v8a/libjniHDHRLib.so
[...]

PR #447, opened 2020-12-30
Use reflection for PointerBufferPoolMXBean to avoid warnings on Android

tmm1:
cc https://github.com/bytedeco/javacpp/issues/446#issuecomment-752294063

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/Pointer.java
+++ b/src/main/java/org/bytedeco/javacpp/Pointer.java
@@ -515,8 +515,8 @@ public class Pointer implements AutoCloseable {
                 Class c = Class.forName("org.bytedeco.javacpp.tools.PointerBufferPoolMXBean");
                 Method method = c.getDeclaredMethod("register");
                 method.invoke(null);
-            } catch (ReflectiveOperationException t) {
-                logger.warn("Could not load PointerBufferPoolMXBean: " + t);
+            } catch (ReflectiveOperationException e) {
+                logger.warn("Could not register PointerBufferPoolMXBean: " + e);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


src/main/java/org/bytedeco/javacpp/Pointer.java
```diff
@@ -512,7 +511,13 @@ public static long parseBytes(String string, long relativeMultiple) throws Numbe
 
         String mx = System.getProperty("org.bytedeco.javacpp.mxbean", "false").toLowerCase();
         if (mx.equals("true") || mx.equals("t") || mx.equals("")) {
-            PointerBufferPoolMXBean.register();
+            try {
+                Class c = Class.forName("PointerBufferPoolMXBean");
```
tmm1, line 515: Does this need to be fully qualified?

```suggestion
                Class c = Class.forName("org.bytedeco.javacpp.tools.PointerBufferPoolMXBean");
```

saudet: Yes, it won't work otherwise.

src/main/java/org/bytedeco/javacpp/Pointer.java
```diff
@@ -512,7 +511,13 @@ public static long parseBytes(String string, long relativeMultiple) throws Numbe
...
+                Method method = c.getDeclaredMethod("register");
+                method.invoke(null);
+            } catch (Throwable t) {
```
saudet, line 518: ReflectiveOperationException would probably suffice here.

tmm1:
```suggestion
            } catch (ReflectiveOperationException t) {
```

matteodg (approved):

PR #456, opened 2021-02-05
Added a flag to prevent system load library

abbasvalliani:
Currently, even if java.library.path is set to nothing, the System.loadLibrary is called which searches through all paths to load the libraries. This creates a significant delay in startup. The pull will allow a switch to control if System.loadLibrary is called and the fix is backwards compatible to prevent breaking existing codecase.

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/Loader.java
+++ b/src/main/java/org/bytedeco/javacpp/Loader.java
@@ -916,10 +916,19 @@ public class Loader {
     static boolean canCreateSymbolicLink = true;
 
     static boolean pathsFirst = false;
+
+    static boolean systemLoadLibrary = true;
+    
     static {
         String s = System.getProperty("org.bytedeco.javacpp.pathsfirst", "false").toLowerCase();
         s = System.getProperty("org.bytedeco.javacpp.pathsFirst", s).toLowerCase();
-        pathsFirst = s.equals("true") || s.equals("t") || s.equals("");
+	pathsFirst = s.equals("true") || s.equals("t") || s.equals("");
+
+	/* Avoids System.loadLibrary load which searches through allow
+	 * all paths and delays overall load time */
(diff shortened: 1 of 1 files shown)


saudet:
Why add a new switch for this? It sounds to me like it's a bug that "java.library.path" doesn't get respected, and we kind need to patch around Java SE for not doing this anyway. BTW, the only reason it's calling System.loadLibrary() is because mainly Android doesn't like System.load() to be called...

saudet:
Actually, I'm not sure that's a bug at all. Java specs make it very clear that we shouldn't change, for example, environment variables, and this is probably why it doesn't respect changes in "java.library.path" after startup. It should be set correctly at startup time.

abbasvalliani:
Understood. Will update it.

saudet:
Seems to be working as intended the way it is too:
```java
public class TestLoad {
        System.loadLibrary("openblas");
    }
}
```
```bash
$ java TestLoad
$ java -Djava.library.path= TestLoad
Exception in thread "main" java.lang.UnsatisfiedLinkError: no openblas in java.library.path
	at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1860)
	at java.lang.Runtime.loadLibrary0(Runtime.java:870)
	at java.lang.System.loadLibrary(System.java:1124)
	at TestLoad.main(TestLoad.java:3)
```

abbasvalliani:
Run it in windows with debug on. 

`Loader.load(opencv_java.class);
`
You'll see searches through the entire paths to find the libraries. Look, I think there is a major issue here. You will always have items in PATH in windows. For each missing library, you're searching the PATH folders which is incredibly slow and takes around 10 seconds plus to load OpenCV. This really slows the entire development process down. There must be a way to turnoff loadLibrary otherwise this is a rather tedious process.

saudet:
I agree, we need a way to disable it, and setting java.library.path works just fine. If you have a version of Java or of Windows where that doesn't work for some reason though, please provide more details about everything that we need to reproduce the issue, and I'll test that see what we can do about it.

PR #457, opened 2021-02-05
Prevent JVM System.loadLibrary call that looks through all paths

abbasvalliani:
Currently, even if java.library.path is set to nothing, the System.loadLibrary is called which searches through all paths to load the libraries. This creates a significant delay in startup.

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/Loader.java
+++ b/src/main/java/org/bytedeco/javacpp/Loader.java
@@ -916,10 +916,19 @@ public class Loader {
     static boolean canCreateSymbolicLink = true;
 
     static boolean pathsFirst = false;
+
+    static boolean systemLoadLibrary = true;
+    
     static {
         String s = System.getProperty("org.bytedeco.javacpp.pathsfirst", "false").toLowerCase();
         s = System.getProperty("org.bytedeco.javacpp.pathsFirst", s).toLowerCase();
-        pathsFirst = s.equals("true") || s.equals("t") || s.equals("");
+	pathsFirst = s.equals("true") || s.equals("t") || s.equals("");
+
+	/* Avoids System.loadLibrary load which searches through allow
+	 * all paths and delays overall load time */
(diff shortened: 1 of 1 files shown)


saudet:
It doesn't do that to start with, this is not an issue: https://github.com/bytedeco/javacpp/pull/456#issuecomment-773747254

saudet:
Please let me know how to reproduce the initial issue. It's not worth merging this if everything works as expected already.

abbasvalliani:
Clearly their is an issue in windows. Mac and Linux is perfectly fine. Will create a simple example tomorrow and figure it out. 
On 2/11/2021 7:18:50 PM, Samuel Audet <notifications@github.com> wrote:
Please let me know how to reproduce the initial issue. It's not worth merging this if everything works as expected already.

saudet:
I can't reproduce this on my installation of  Windows 10 either. This is what I get with libopenblas.dll from MSYS2 in the PATH:
```java
public class TestLoad {
        System.loadLibrary("libopenblas");
    }
}
```
```bash
$ java -version
java version "1.8.0_152"
Java(TM) SE Runtime Environment (build 1.8.0_152-b16)
Java HotSpot(TM) 64-Bit Server VM (build 25.152-b16, mixed mode)

$ java TestLoad

$ java -Djava.library.path= TestLoad
Exception in thread "main" java.lang.UnsatisfiedLinkError: no libopenblas in java.library.path
        at java.lang.ClassLoader.loadLibrary(ClassLoader.java:1867)
        at java.lang.Runtime.loadLibrary0(Runtime.java:870)
        at java.lang.System.loadLibrary(System.java:1122)
        at TestLoad.main(TestLoad.java:3)
```

saudet:
BTW, if you've only tried it with `mvn -Djava.library.path= ...`, it might be because Maven sets the system property too late.
Try instead setting the environment variable `MAVEN_OPTS=-Djava.library.path=`.

PR #459, opened 2021-02-07
create a pretty format pom.xml

Celebrate-future:


saudet:
Does this only change the formatting? Or are there other changes in here?

Celebrate-future:
> Does this only change the formatting? Or are there other changes in here?

Only change the formatting.

saudet:
Ok, the formatting is fine as it is. No need to change it.

PR #461, opened 2021-02-14
Add `Info.readOnlyMembers` to disable generating setters for publi…

equeim:
Sometimes it is needed to pass data from C++ to Java in form of struct, but there is no need to make it mutable. In this case, generated setter methods are unused and only increase size of native library and compiled Java classes.

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/tools/Context.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Context.java
@@ -43,10 +43,10 @@ class Context {
         javaName = c.javaName;
         constName = c.constName;
         constBaseName = c.constBaseName;
+        immutable = c.immutable;
         inaccessible = c.inaccessible;
         objectify = c.objectify;
         virtualize = c.virtualize;
-        readOnlyMembers = c.readOnlyMembers;
         variable = c.variable;
         infoMap = c.infoMap;
         templateMap = c.templateMap;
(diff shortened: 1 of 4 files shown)


src/main/java/org/bytedeco/javacpp/tools/Context.java
```diff
@@ -62,6 +63,7 @@
     boolean inaccessible = false;
     boolean objectify = false;
     boolean virtualize = false;
+    boolean readOnlyMembers = false;
```
saudet, line 66: For consistency, what about naming this "constify"? (It's actually a term that C++ programmers use!)

saudet: Or maybe just "immutable"?

equeim: But that way this property would also have to disable non-const C++ methods (and maybe some other things too) otherwise it would be confusing.

Although, current name also doesn't make it 100% clear that it deals only with data members, since methods in C++ are technically called member functions.

saudet: Right w.r.t to "const", but "immutable" can only refer to *data* and usually only to "properties", even in the case of C++. Functions, or even variables for that matter, are not "properties", because well we can't do things like make them immutable. :)

saudet: I see, we can't currently apply this to individual fields, but ideally it's something that it should let us do. I guess we can wait until someone asks for it before implementing this though.

Issue #464, opened 2021-03-04
Passing std::vector<Foo>& to a method when Foo does not have a public default constructor


magzot:
Trying to generate jni wrappers for a c++ library that has the pattern of creating objects and pushing them into a std::vector that is passed in byref to a factory method. And the types do not have default constructors. 

This is how it would look in c++

```
std::vector<Box> boxes;
BoxFactory factory;
factory.CreateBoxes(10, boxes);
```

And to use the generated java it would look something like this
```
JniBox.BoxFactory factory = new JniBox.BoxFactory();
JniBox.Box box = new JniBox.Box(); // <<-- This is private in c++
//JniBox.Box box = new JniBox.Box(new Pointer()); //Would this work?
factory.CreateBoxes(10, box);
```

Problem is that the Box type does not have a public default constructor so does not seem to be possible to call this factory method using the generated java code.

I have attached a full example including c++ and java code.

[javacpp_example.zip](https://github.com/bytedeco/javacpp/files/6081945/javacpp_example.zip)

saudet:
It sounds like we don't need to create a `std::vector<Box>` from Java, so this shouldn't be a problem. Passing a `new Box(null)` for the output argument should indeed just work. It doesn't?

magzot:
ah, you're right :) that just works! wasn't really sure how to use the Pointer types. But, it does not work with the real scenario... though, so guess something else is going wrong there. What's happening is that the "limit" field of the Pointer does now grow. So seems like nothing is added to the vector, or the VectorAdapter does not pick it up some how. But maybe I should compare the generated c++ code for this example and the real scenario and see if I can understand what is going wrong. 

I guess you can close this issue then if you like :)

saudet:
Well, let me know if you can update it to reflect the real scenario. Thanks!

magzot:
Hi! I have modified my example now so that the generated jni methods in the example and the real lib now look the same (except for names). Also I added an operator=() to the Box class and i think that maybe I found something. Seems like the operator=() is called on an object that has been deleted! When I did not have the operator=() the copy constructor was called instead. With the real lib the assignment operator is doing things that most often is causing an immediate segfault when called on a deleted object. 

Attached new example. 

[javacpp_example.zip](https://github.com/bytedeco/javacpp/files/6097355/javacpp_example.zip)

saudet:
[...]

PR #465, opened 2021-03-08
Use uninitialized_copy to copy into uninitialized memory

magzot:
Use `uninitialized_copy` to prevent from calling `copy assignment operator=()` on uninitialised memory (unconstructed `this`).


magzot:
resolves #464

saudet:
I see, this essentially allows to guarantee that the copy constructor always gets called instead of `operator=()`. Nice catch!

Issue #483, opened 2021-05-28
Javacpp does not export package required by presets


reckart:
Presets such as the openblas preset in recent versions inherit from `org.bytedeco.javacpp.presets.javacpp`:

```
@Properties(inherit = javacpp.class, global = "org.bytedeco.openblas.global.openblas_nolapack", value = {
...
public class openblas_nolapack implements LoadEnabled, InfoMapper {
```

In an OSGI context, this leads to a ClassNotFoundException being thrown as part of loading the class annotations:

```
java.lang.ClassNotFoundException: org.bytedeco.javacpp.presets.javacpp cannot be found by <MY-BUNDLE>
```

I only discover the above exception by putting a breakpoint in the constructor of the `TypeNotPresentExceptionProxy`...
Otherwise, here is a stack trace showing the annotation loading:

```
Thread [AsyncThread-1] (Suspended (breakpoint at line 41 in TypeNotPresentExceptionProxy))	
	owns: Nd4j  (id=180)	
	owns: ConcurrentHashMap$ReservationNode<K,V>  (id=181)	
	TypeNotPresentExceptionProxy.<init>(String, Throwable) line: 41	
	AnnotationParser.parseClassValue(ByteBuffer, ConstantPool, Class<?>) line: 429	
	AnnotationParser.parseClassArray(int, ByteBuffer, ConstantPool, Class<?>) line: 724	
	AnnotationParser.parseArray(Class<?>, ByteBuffer, ConstantPool, Class<?>) line: 531	
	AnnotationParser.parseMemberValue(Class<?>, ByteBuffer, ConstantPool, Class<?>) line: 355	
	AnnotationParser.parseAnnotation2(ByteBuffer, ConstantPool, Class<?>, boolean, Class<Annotation>[]) line: 286	
	AnnotationParser.parseAnnotations2(byte[], ConstantPool, Class<?>, Class<Annotation>[]) line: 120	
	AnnotationParser.parseAnnotations(byte[], ConstantPool, Class<?>) line: 72	
	Class<T>.createAnnotationData(int) line: 3521	
	Class<T>.annotationData() line: 3510	
	Class<T>.createAnnotationData(int) line: 3526	
	Class<T>.annotationData() line: 3510	
	Class<T>.createAnnotationData(int) line: 3526	
	Class<T>.annotationData() line: 3510	
	Class<T>.getAnnotation(Class<A>) line: 3415	
	Loader.checkPlatform(Class<?>, Properties) line: 1010	
[...]

PR #490, opened 2021-06-02
#483 - Javacpp does not export package required by presets

reckart:
- Add package-info with OSGI annotation so that the presets packages is exported for use by e.g. the openblas preset
- Add m2e lifecycle bindings for the maven-plugin-plugin to avoid error in Eclipse when importing the project

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 
+ * Add missing `presets/package-info.java` required for OSGi and add profile for M2Eclipse ([pull #490](https://github.com/bytedeco/javacpp/pull/490))
  * Remove unnecessary mutex lock for pthreads on callbacks in `Generator` ([pull #489](https://github.com/bytedeco/javacpp/pull/489))
  * Fix `@AsUtf16` handling for setter methods paired with getters in `Generator` ([pull #488](https://github.com/bytedeco/javacpp/pull/488))
  * Allow defining `NO_JNI_DETACH_THREAD` to avoid overhead from pthreads on callbacks ([issue #486](https://github.com/bytedeco/javacpp/issues/486))
--- a/pom.xml
+++ b/pom.xml
@@ -581,6 +581,72 @@ Import-Package: \
         </plugins>
       </build>
     </profile>
+    
+    <!-- *********************************** -->
+    <!-- * Eclipse: m2e lifecycle bindings * -->
+    <!-- *********************************** -->
+    <profile>
+      <id>m2e</id>
+      <activation>
+        <property>
+          <name>m2e.version</name>
+        </property>
+      </activation>
+      <build>
+        <pluginManagement>
+          <plugins>
+            <!--This plugin's configuration is used to store Eclipse m2e settings 
+                only. It has no influence on the Maven build itself. -->
+            <plugin>
+              <groupId>org.eclipse.m2e</groupId>
+              <artifactId>lifecycle-mapping</artifactId>
+              <version>1.0.0</version>
+              <configuration>
+                <lifecycleMappingMetadata>
+                  <pluginExecutions>
+                    <!-- *********************************************** -->
+                    <!-- The Maven Dev Connector for Eclipse m2e is no   -->
+                    <!-- longer maintained. We copy the relevant part    --> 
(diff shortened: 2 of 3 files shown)


reckart:
I made the change according to the pointer provided by @timothyjward

reckart:
Building the package, I can now see this in the MANIFEST.MF:

```
Export-Package: org.bytedeco.javacpp;uses:="org.bytedeco.javacpp.annot
 ation,org.bytedeco.javacpp.presets";version="1.5.6",org.bytedeco.java
 cpp.annotation;version="1.5.6",org.bytedeco.javacpp.indexer;uses:="or
 g.bytedeco.javacpp";version="1.5.6",org.bytedeco.javacpp.presets;uses
 :="org.bytedeco.javacpp,org.bytedeco.javacpp.annotation";version="1.5
 .6",org.bytedeco.javacpp.tools;uses:="javax.management";version="1.5.
 6"
```

So it looks like the `org.bytedeco.javacpp.presets` package is exported now.

saudet:
@timothyjward Does this look alright to you?

reckart:
@agibsonccc if you do a follow-up release to DL4J 1.0.0-M1 soon, it would be great if this change could make it in.

saudet:
It's probably not going to be enough. I think we need to add those package-info.java and what not to the presets themselves first.

reckart:
At least in our case, we pack the presets (e.g. openblas.jar) into our own OSGI bundle which then imports the packages from JavaCPP. 

We do not add JavaCPP directly into our bundle because it already is a bundle and bundling bundles doesn't really seem to be a good idea.

Do you plan to turn the presets JARs into OSGI bundles as well?

agibsonccc:
@reckart I don't see why not. It would depend on @saudet 's release schedule. I have 1 pending PR that depends on a javacpp upgrade that I can't merge yet, but would be happy to otherwise. A javacpp version bump doesn't usually cost us anything.

saudet:
> At least in our case, we pack the presets (e.g. openblas.jar) into our own OSGI bundle which then imports the packages from JavaCPP.
> 
> We do not add JavaCPP directly into our bundle because it already is a bundle and bundling bundles doesn't really seem to be a good idea.
> 
> Do you plan to turn the presets JARs into OSGI bundles as well?

Ideally, yes, contributions are welcome. If you're bundling everything anyway, you should also bundle JavaCPP and ignore it's own bundle though.

reckart:
We had trouble integrating JavaCPP into the bundle in the past. As far as I remember, the Maven bundle plugin somehow merged some of the OSGI metadata from JavaCPP into the main bundle metadata which messed things up. At that point, we decided to handle JavaCPP separately. Also because in perspective, we'd have multiple bundles with native libraries and they should probably all share the same JavaCPP because as far as I understand Java native library loading should be centralized under a single classloader to avoid issues with attempts of loading the same native library multiple times. Not sure if I explain this well - it's a bit if a difficult terrain.

So having everything in one bundle is only a temporary convenience for us that hopefully we'll move away from sooner or later.

timothyjward:
> We had trouble integrating JavaCPP into the bundle in the past. As far as I remember, the Maven bundle plugin somehow merged some of the OSGI metadata from JavaCPP into the main bundle metadata which messed things up. At that point, we decided to handle JavaCPP separately. Also because in perspective, we'd have multiple bundles with native libraries and they should probably all share the same JavaCPP because as far as I understand Java native library loading should be centralized under a single classloader to avoid issues with attempts of loading the same native library multiple times. Not sure if I explain this well - it's a bit if a difficult terrain.
> 
> So having everything in one bundle is only a temporary convenience for us that hopefully we'll move away from sooner or later.

This is correct - you should only embed things that are designed to be completely internal to your bundle/module (much like using the maven-shade-plugin). In this case there are multiple different bundles using JavaCPP and (unless DL4J has changed its structure quite a bit) the fact that those modules are using JavaCPP is not an implementation detail. 

As a general rule, if a bundle which embeds some packages exports those packages then it probably shouldn't be embedding them.

agibsonccc:
@timothyjward  not a lot has changed. We're still sitting on your original pull request that adds osgi support and I think the same blockers are still there.

PR #510, opened 2021-08-04
Add parameter classes to Cache mojo

HGuillemet:
When using 
```
mvn org.bytedeco:javacpp:cache -Djavacpp.platform=linux-x86_64
```
from a project using, e.g., opencv, the cache plugin walks through all classes in the opencv presets and attempts to load the javacpp-annotated ones.
This will try to load and cache a lot of libraries, potentially from other presets like cpython, while the project may use only a subset of them.
This PR adds a `classes` parameter to limit the scope of the class walk:
```
mvn -Dclasses=org.bytedeco.opencv.global.opencv_core,org.bytedeco.opencv.global.opencv_imgproc org.bytedeco:javacpp:cache -Djavacpp.platform=linux-x86_64
```
Does it make sense ?

There is probably smarter but more complex alternatives, like using jdeps or similar to automatically cache what is actually used by the project.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,6 @@
 
+ * Add `classOrPackageNames` parameter to `CacheMojo` ([pull #510](https://github.com/bytedeco/javacpp/pull/510))
+
 ### August 2, 2021 version 1.5.6
  * Add missing export to `module-info.java` file for presets package ([pull #508](https://github.com/bytedeco/javacpp/pull/508))
  * Add `@NoException(true)` value to support overriding `virtual noexcept` functions
--- a/src/main/java/org/bytedeco/javacpp/tools/BuildMojo.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/BuildMojo.java
@@ -238,7 +238,7 @@ public class BuildMojo extends AbstractMojo {
     @Parameter(defaultValue = "${plugin}", required = true, readonly = true)
     PluginDescriptor plugin;
 
-    String[] merge(String[] ss, String s) {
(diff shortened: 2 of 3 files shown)


saudet:
Sure, that makes sense, but let's name the property "javacpp.classOrPackageNames" to make it consistent with BuildMojo, and be aware that it's not possible to populate arrays like that with system properties on the command line or elsewhere... In any case, please fix the formatting! Thanks

HGuillemet:
Fixed formatting and property name.
What do you mean by it's not possible to populate arrays ? It seems to work for me:
```
$ mvn -Djavacpp.classOrPackageNames=org.bytedeco.opencv.global.opencv_core,org.bytedeco.opencv.global.opencv_imgproc org.bytedeco:javacpp:1.5.7-SNAPSHOT:cache -Djavacpp.platform=linux-x86_64
...
Info: Loading class org.bytedeco.opencv.global.opencv_core
Info: Loading class org.bytedeco.opencv.global.opencv_imgproc
```

saudet:
Ok, that must be a new feature of Maven I'm not aware of... Let's also put a single String property "javacpp.classOrPackageName" at least for older versions of Maven and to be consistent with Builder?

HGuillemet:
Done.
Some properties are prefixed with the fqdn like `org.bytedeco.javacpp.cachedir`. What is the logic behind the naming ?

saudet:
The system properties with long names are typically used at runtime, when many other libraries may be loaded, with their own system properties that we can't control. Maven plugins on the other hand are typically not so numerous in any given build and something we can also control, so to reduce typing on the command line, the names of their properties are typically short, that's all.

Issue #506, opened 2021-07-14
Integration of JavaCPP fails for Eclipse/OSGi application


HannesWell:
I'm trying to integrate bytedeco/cpython into an Eclipse based application (which uses the Equinox OSGi runtime, like all other Eclipse applications too). Even tough I tried multiple different approaches to integrate cpython and javacpp, but non of them was successful:
- include it via the Target-Platform
- embed the main jars directly into a Eclipse plugin and the platform specific jars in corresponding fragments
- embed the unpacked content of the jars in corresponding Eclipse plug-ins/fragments

All suffered the same problem: the `javacpp.Loader` cannot handle URL's using the `bundleresource` schema (like `bundleresource://491.fwk108568334:2/org/bytedeco/javacpp/windows-x86_64/api-ms-win-crt-locale-l1-1-0.dll`) properly.
These URLs are returned by the EquinoxClassLoader used by Eclipse when the `javacpp.Loader` quires resources in its `findResources()` methods.
With those URLs the caching of packages fails and the libraries cannot be loaded.

Therefore my question is: Is it possible to use javacpp and its preset in an Eclipse/OSGi application? And if yes are there some best-practices how to integrate them?
I noticed you have a `osgi` branch where some contributions were made some time ago, but it seems to be stale and I could not find information how to use the current main artefacts in Eclipse/OSGi.

If it is not possible yet, one solution to solve the problem mentioned above could be to include org.eclipse.platform:org.eclipse.osgi as optional dependency and use the `org.eclipse.osgi.storage.url.BundleURLConverter` to resolve `bundleresource` URLs if that class is available at runtime (if it cannot be loaded it is likely that it is not required).

If the suggested change is suitable and wanted, I can create a PR for it.

saudet:
The presets currently don't come with anything for OSGi, so users like @reckart simply create their own uber JAR bundle out of whatever they need, and it works fine.

If you would like to contribute changes to enhance OSGi support, that would be great! Thanks

/cc @timothyjward

reckart:
It works for me to:

* install JavaCPP as bundles via the target platform (1.9.4 works currently best)
* install Apache Aries SPI loader via the target platform
* package up any jar libs and native libs (e.g. DL4J) into a fat bundle

HannesWell:
[...]

PR #511, opened 2021-08-17
Support directory extraction in modular OSGi runtimes (issue #506)

HannesWell:
In order to resolve issue #506 I implemented a solution based on @timothyjward suggestion(https://github.com/bytedeco/javacpp/issues/506#issuecomment-886551437) to support extraction of directories in case javacpp is a bundle in a OSGi runtime.

The indention of this solution is to be independent of a specific OSGi framework implementation.

In contrast to @timothyjward suggestion I used `Bundle.findEntries()` instead of `BundleWiring.listResources()` because it is a bit less code (no need to obtain the BundleWiring) and already gives the desired URLs of all files in the directory and not only the String paths within the bundle (avoids 'manual' URL construction).
Since the javacpp bundles don't have a Bundle-Classpath header specified in their Manifest it defaults to the dot ".", so the bundle-classpath is just the jar content. And since the search for resources is limited to local resources, I think the result should always be the same. In Equinox findEntries() returns a URL using the bundleentry schema instead of bundleresources, but that's not an issue.

I have tested it with Equinox and the cpython javacpp preset and it works quite well.

The only flaw is that the javacpp windows-dll's (I assume it is the same case for linux) are cached twice:
1. in the folder for the `javacpp` bundle
2. in the folder of the `cpython` bundle

The second location is obviously not wanted. I tried to investigate that and I think the reason is that the dll's are loaded a second time in the context of a class of the cpython bundle/preset in `Loader.load(Class, Properties, boolean, String)` when `oldUrls == null || urls.length > 0` is true and urls is not empty.

I think this problem could be avoided if we were able to obtain the bundle from the resource-URLs directly. Then it would be possible to find the containing bundle. But unfortunately I don't know a proper official/API way to do this. If any of you knows one, it is more than welcome.

The only approach I found, that would hopefully work with different OSGi implementations, was to maintain a map of URL-hosts to the corresponding bundle/bundle-id that is filled when the Loader is initialized and also keeps track of installed/uninstalled bundles via a BundleListener.
At least in Equinox the resource-URL's host is always the same for the same bundle. Do you know if this is the case for other OSGi implementations as well?
This as the additional advantage that there is no need to pass a class-argument all the way down. I can submit this approach with separate commit to this PR.

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/Loader.java
+++ b/src/main/java/org/bytedeco/javacpp/Loader.java
@@ -524,22 +524,22 @@ public class Loader {
             if (!noSubdir) {
                 cacheSubdir = new File(cacheSubdir, urlFile.getParentFile().getName());
             }
-		} else if (OSGiBundleResourceLoader.isOSGiRuntime()) {
-			// TODO: what happens if this is another URL in a OSGi environment?
-			// I think it is unlikely that is called with URL-schema?!
-			if (!noSubdir) {
-				String subdirName = OSGiBundleResourceLoader.getContainerBundleName(resourceURL);
-				if (subdirName != null) {
-					String parentName = urlFile.getParentFile().toString();
-					if (parentName != null) {
-						subdirName = subdirName + File.separator + parentName;
-					}
-					cacheSubdir = new File(cacheSubdir, subdirName);
-				}
-				size = urlConnection.getContentLengthLong();
-				timestamp = urlConnection.getLastModified();
-			}
-		} else {
+        } else if (OSGiBundleResourceLoader.isOSGiRuntime()) {
+            // TODO: what happens if this is another URL in a OSGi environment?
+            // I think it is unlikely that is called with URL-schema?!
+            if (!noSubdir) {
+                String subdirName = OSGiBundleResourceLoader.getContainerBundleName(resourceURL);
+                if (subdirName != null) {
+                    String parentName = urlFile.getParentFile().toString();
+                    if (parentName != null) {
+                        subdirName = subdirName + File.separator + parentName;
+                    }
+                    cacheSubdir = new File(cacheSubdir, subdirName);
+                }
+                size = urlConnection.getContentLengthLong();
+                timestamp = urlConnection.getLastModified();
+            }
+        } else {
             if (urlFile.exists()) {
                 size = urlFile.length();
                 timestamp = urlFile.lastModified();
--- a/src/main/java/org/bytedeco/javacpp/Loader.java
+++ b/src/main/java/org/bytedeco/javacpp/Loader.java
@@ -766,7 +766,6 @@ public class Loader {
         if (urlConnection instanceof JarURLConnection) {
             JarFile jarFile = ((JarURLConnection) urlConnection).getJarFile();
             JarEntry jarEntry = ((JarURLConnection) urlConnection).getJarEntry();
-            String jarFileName = jarFile.getName();
             String jarEntryName = jarEntry.getName();
             if (!jarEntryName.endsWith("/")) {
                 jarEntryName += "/";
@@ -783,8 +782,7 @@ public class Loader {
                         File file = new File(directoryOrFile, entryName.substring(jarEntryName.length()));
                         if (entry.isDirectory()) {
                             file.mkdirs();
-                        } else if (!cacheDirectory || !file.exists() || file.length() != entrySize
-                                || file.lastModified() != entryTimestamp || !file.equals(file.getCanonicalFile())) {
+                        } else if (!cacheDirectory || isCacheFileCurrent(file, entrySize, entryTimestamp)) {
                             // ... extract it from our resources ...
                             file.delete();
                             String s = resourceURL.toString();
@@ -812,8 +810,7 @@ public class Loader {
                     File file = new File(directoryOrFile, entryName.substring(directoryName.length()));
                     if (entryName.endsWith("/")) { // is directory
                         file.mkdirs();
-                    } else if (!cacheDirectory || !file.exists() || file.length() != entrySize
-                            || file.lastModified() != entryTimestamp || !file.equals(file.getCanonicalFile())) {
+                    } else if (!cacheDirectory || isCacheFileCurrent(file, entrySize, entryTimestamp)) {
                         // ... extract it from our resources ...
                         file.delete();
                         // FIXME: check if directories have to be extracted again?!
@@ -877,6 +874,11 @@ public class Loader {
             }
         }
         return file;
+    }
+
+    private static boolean isCacheFileCurrent(File file, long entrySize, long entryTimestamp) throws IOException {
+        return !file.exists() || file.length() != entrySize || file.lastModified() != entryTimestamp
+                || !file.equals(file.getCanonicalFile());
     }
 
     /** Returns {@code findResources(cls, name, 1)[0]} or null if none. */
--- a/src/main/java/org/bytedeco/javacpp/Loader.java
+++ b/src/main/java/org/bytedeco/javacpp/Loader.java
@@ -55,12 +55,16 @@ import java.util.Properties;
 import java.util.WeakHashMap;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
+
 import org.bytedeco.javacpp.annotation.Cast;
 import org.bytedeco.javacpp.annotation.Name;
 import org.bytedeco.javacpp.annotation.Platform;
 import org.bytedeco.javacpp.annotation.Raw;
 import org.bytedeco.javacpp.tools.Builder;
 import org.bytedeco.javacpp.tools.Logger;
+import org.osgi.framework.Bundle;
+import org.osgi.framework.FrameworkUtil;
+import org.osgi.framework.Version;
 
 /**
  * The Loader contains functionality to load native libraries, but also has a bit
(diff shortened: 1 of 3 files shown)


timothyjward:
> At least in Equinox the resource-URL's host is always the same for the same bundle. Do you know if this is the case for other OSGi implementations as well?

The guarantees about the URL scheme are described in the specification. You will need to be careful using the host as this sometimes includes information about the bundle revision which changes when the bundle is refreshed or updated.

The most relevant section of the specification is here:
http://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#i3178562

saudet:
Like I said before, I'd like it if we could do everything with reflection, but if this is not desirable, another way of going about it is to put everything related to OSGi in another class, just like stuff related to MXBeans isn't in Pointer, because it's not available on Android, but in tools/PointerBufferPoolMXBean. In this case, I'd see something like maybe tools/OsgiBundleResourceLoader as per Spring?
https://docs.spring.io/spring-osgi/docs/current/api/org/springframework/osgi/io/OsgiBundleResourceLoader.html

Also doing without having to pass around the Class object everywhere would be great, yes. Thanks for looking into this!

HannesWell:
I have move all code that directly interacts with OSGi APIs into tools/OSGiBundleResourceLoader and also extract the bundle from the URL.
There are still some TODOs and time measurements that I will resolve/remove once we agreed on the code.

> > At least in Equinox the resource-URL's host is always the same for the same bundle. Do you know if this is the case for other OSGi implementations as well?
> 
> The guarantees about the URL scheme are described in the specification. You will need to be careful using the host as this sometimes includes information about the bundle revision which changes when the bundle is refreshed or updated.
> 
> The most relevant section of the specification is here:
> http://docs.osgi.org/specification/osgi.core/7.0.0/framework.module.html#i3178562

Thanks for your remark. As you said the OSGi spec does not specify any requirements on the URLs host and I did not find a better way to extract the bundle a URL refers to. I tried to work around a potentially changing host by attaching Bundle/FrameworkListeners but I had no immediate test case yet because our Eclipse application is quite static at the point cpython is used.
So @timothyjward what do you think about this approach? Could it work?

Of course it would be much simpler if there were a OSGi service that provides the bridge from the URL to the bundle or e.g. if the connections created for such URLs would implement `BundleReference` like the ClassLoaders do.

HannesWell:
@timothyjward: Furthermore, is there a deeper meaning why you suggested `BundleWiring.listResources()` over `findEntries()`? You have mentioned both.

timothyjward:
> @timothyjward: Furthermore, is there a deeper meaning why you suggested `BundleWiring.listResources()` over `findEntries()`? You have mentioned both.

The two methods do different things. 

`findEntries()` Is a simple way to access a path within the bundle, much like getting a JarEntry from a JarFile. `findEntries()` ignores any internal classpath in the bundle, and therefore does not require the bundle to be resolved.

`listResources()` is different in that it uses the Bundle Classloader to find the resource. This therefore honours the bundle classpath, and therefore may not be able to locate some things that can be found by `findEntries()`. On the other hand as `listResources()` uses the classloader it can also see things imported from other bundles. 

The main reason that you might choose one over the other is that the native code is loaded using a  Class as context. It’s therefore pretty clear to me that the classpath is supposed to be taken into account, and the library should be located relative to the classpath (e.g. in the same package as the class). This would point me to listResources as the better candidate. You would probably want to pass the flag which tells it not to return resources from outside the bundle though.

src/main/java/org/bytedeco/javacpp/tools/OSGiBundleResourceLoader.java
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	private static class OSGiEnvironmentLoader {
+		// Code using OSGi APIs has to be encapsulated into own class
+		// to prevent NoClassDefFoundErrors in OSGi environments
+
+		private static void initialize() {
+			BundleContext context = getBundleContext();
+			if (context != null) {
+				indexAllBundles(context);
```
timothyjward: This code could be simplified by using a BundleTracker. It would also fix the race condition where a bundle is added between the index call and adding the listener.

src/main/java/org/bytedeco/javacpp/tools/OSGiBundleResourceLoader.java
```diff
@@ -0,0 +1,178 @@
...
+		private static void indexAllBundles(BundleContext context) {
+			if (context != null) {
+				for (Bundle bundle : context.getBundles()) {
+					HOST_2_BUNDLE.put(getBundleURLHost(bundle), bundle.getBundleId());
+				}
+			}
+		}
+
+		private static String getBundleURLHost(Bundle bundle) {
+			return bundle.getEntry("/").getHost();
```
timothyjward: This works with getEntry, but not with getResource as there can be multiple entries on the bundle classpath.

src/main/java/org/bytedeco/javacpp/Loader.java
```diff
@@ -771,14 +789,43 @@ public static File extractResource(URL resourceURL, File directoryOrFile,
...
                 }
+                System.out.println("Extract took " + (System.currentTimeMillis() - start) + "ms, for:" + resourceURL);
                 return directoryOrFile;
             }
         }
+        if (OSGiBundleResourceLoader.isOSGiRuntime()) {
+            Enumeration<URL> directoryEntries = OSGiBundleResourceLoader.getBundleDirectoryContent(resourceURL);
+            if (directoryEntries != null && directoryEntries.hasMoreElements()) { // a not empty directory
+                String directoryName = resourceURL.getPath();
+                while (directoryEntries.hasMoreElements()) {
```
saudet, line 806: This while loop looks pretty similar to the one above. Is the idea that some OSGi implementations may support other things that JAR files?

src/main/java/org/bytedeco/javacpp/tools/OSGiBundleResourceLoader.java
HannesWell: Thanks for that hint. I applied it.

src/main/java/org/bytedeco/javacpp/Loader.java
HannesWell: This is intentionally. Actually I would have liked to move the extraction part into an own method. But because in case of a JarURLConnection the URL is created just before extractResource is called again this is not possible without creating the URL before (and therefore also in cases of a directory.

And yes a bundle can be a jar but also can be extracted into a directory (which is for example very common if you develop Eclipse plug-ins and use them in a debugged application started from the IDE).

HannesWell:
> 
> 
> > @timothyjward: Furthermore, is there a deeper meaning why you suggested `BundleWiring.listResources()` over `findEntries()`? You have mentioned both.
> 
> The two methods do different things.
> 
> `findEntries()` Is a simple way to access a path within the bundle, much like getting a JarEntry from a JarFile. `findEntries()` ignores any internal classpath in the bundle, and therefore does not require the bundle to be resolved.
> 
> `listResources()` is different in that it uses the Bundle Classloader to find the resource. This therefore honours the bundle classpath, and therefore may not be able to locate some things that can be found by `findEntries()`. On the other hand as `listResources()` uses the classloader it can also see things imported from other bundles.
> 
> The main reason that you might choose one over the other is that the native code is loaded using a Class as context. It’s therefore pretty clear to me that the classpath is supposed to be taken into account, and the library should be located relative to the classpath (e.g. in the same package as the class). This would point me to listResources as the better candidate. You would probably want to pass the flag which tells it not to return resources from outside the bundle though.

Thanks for your elaboration. I understand your point regarding bundle native code, but I think in this case it is not applicable because we are just looking for children of a URL pointing to a directory. So we are just interested in the actual content of the directory and not of the 'path' with which they are referenced. When the directory can be found through the classloader I'm pretty sure that all its children can be found as well by concatenating a corresponding sub-path.
And since a URL store the location of a resource and not how it can be found, I think this is equivalent in terms of discoverability/accessibility.
The reason why I would prefer `findEntries()` is that is consistent to how a `JarFile` is searched in a non-OSGi environment, where it is just looked up what's in the jar and not what can be found on the classpath. And it gives me the URL directly which is more convenient.

HannesWell:
Because I'm unsure if the current approach using a `BundleTracker` and `FrameworkListener` catches all possible changes in an URLs host, @timothyjward mentioned that for example in Felix a changed BundleRevision/Wiring might reflect into a changed URL-Host and other implementations might create a completely different host, so there could be more pitfalls I don't observe in my Equinox case because the URL host is simply not specified, and because IMHO the current approach is already quite complicated, I'm thinking of a different approach to get the Bundle of a URL:

In case javacpp runs as a bundle in a OSGi environment the calls to `Class.getResource()` and `ClassLoader.getResources()` could be redirected to the OSGiBundleResourceLoader as well and then the Class(Loader)'s bundle could be saved together with the returned URL in a static WeakHashMap.
This way we don't have to relay on implementation specific behaviour that might or might not be consistent. And as fas as I can oversee it, `Class.getResource()` and `ClassLoader.getResources()` is only called in `Loader.findResources()`, so the redirection is not a big issue. Or are also URLs passed to the loader that are created at other locations?

What do you think?

src/main/java/org/bytedeco/javacpp/Loader.java
```diff
@@ -22,6 +22,10 @@
 
 package org.bytedeco.javacpp;
 
+import static org.bytedeco.javacpp.tools.OSGiBundleResourceLoader.getOSGiClassLoaderResources;
+import static org.bytedeco.javacpp.tools.OSGiBundleResourceLoader.getOSGiClassResource;
+import static org.bytedeco.javacpp.tools.OSGiBundleResourceLoader.isOSGiRuntime;
```
saudet, line 27: Let's try to keep the imports clean of OSGi stuff. Android in particular has a tendency to trip on those things since it tries to load all the classes there, and fails. This also means that we can't put anything that loads those classes by default, so you'll need to revert those changes too.

saudet:
> What do you think?

Are you asking me? It sounds fine, yes, as long as by default it doesn't try to load anything, other than by reflection with fail safe.

Issue #452, opened 2021-01-31
[ffmpeg] Loading native libraries on Windows 10 takes a long time


edwinRNDR:
I am using javacpp/ffmpeg on Windows 10 (2004) and found that loading the native libraries takes longer for Javacpp version 1.5.4 in comparison to 1.5.3. On 1.5.4 it takes roughly 7 seconds to load the libraries, which is in contrast to the 1.5 seconds version 1.5.3 takes.

In the gist below you find timestamped logs of the native libraries being loaded. Without having looked at the loader sources I am surprised to find that more time is spent in loading `api-ms-win-crt-*`

https://gist.github.com/edwinRNDR/676ea1a75548b31edc07cd699010f75e

saudet:
Please try again with 1.5.5-SNAPSHOT: http://bytedeco.org/builds/

edwinRNDR:
I updated the gist with the logs for 1.5.5-SNAPSHOT. My observation is that it takes about the same amount of time to load as 1.5.4, but it seems to report less library load failures.

saudet:
Since FFmpeg is compiled with GCC, it doesn't need those DLLs from MSVC, so you could exclude from your class path the JAR files for `javacpp-platform`.

edwinRNDR:
I am not entirely sure how I should do that, the javacpp platform jars do not seem to be on the classpath?

![image](https://user-images.githubusercontent.com/983309/106386183-07340b80-63d4-11eb-9f69-74e966d6667b.png)

saudet:
So, where are they getting loaded from? Maybe you should add it, and that's the problem.

edwinRNDR:
Adding javacpp-platform brings the loading time down to 2.5 seconds, which is an improvement but it is still consistently a second more than 1.5.3. I updated the gist with the log (It is the file at the bottom)

Adding javacpp-platform to 1.5.3 seems to make it 250ms faster but that could be noise in my measurements.

saudet:
Like I said, use 1.5.5-SNAPSHOT. That's been fixed.

saudet:
Well, this isn't happening here. I've just tested this on my installation of Windows 10, and 4.2.2-1.5.3 takes about 1200 ms, 4.3.1-1.5.4, about 1500 ms, and 4.3.1-1.5.5-SNAPSHOT, about 700 ms. If you figure out what's special about your machine, please let me know!

edwinRNDR:
[...]

PR #512, opened 2021-08-20
[Loader] improve performance of resource extraction and library search

HannesWell:
With this PR I suggest changes to reduce the runtime of resource extraction and library search performed by the javacpp.Loader.

My benchmark is a stripped variant of the embeddedpython.Python that uses only javacpp and cpython (but not numpy) and the 'regular' variant of embeddedpython.Python  but with updated javacpp dependencies.
The benchmark is just the following program, which is absolutly dominated by the initialization of the embeddedpython.Python :
```
long start = System.currentTimeMillis();
Python.put("a", 5);
Python.put("b", 3);
Python.exec("v = a/b");
Object x = Python.get("v");
System.out.println(x);
System.out.println("Took " + (System.currentTimeMillis() - start) + "ms");
```

On my Windows 10 computer this program takes (when all files are already cached in a previous run) with the current javacpp master 2000ms+-20 for the stripped version respectively 3630ms+-20 for the regular version.

With the suggested changes, the stripped version takes around 800+-20ms and the regular version 1400ms+-20.
So this improves the runtime of the initialization by more than the factor of two.

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/Loader.java
+++ b/src/main/java/org/bytedeco/javacpp/Loader.java
@@ -49,12 +49,15 @@ import java.util.Deque;
 import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Properties;
+import java.util.Set;
 import java.util.WeakHashMap;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
+
 import org.bytedeco.javacpp.annotation.Cast;
 import org.bytedeco.javacpp.annotation.Name;
 import org.bytedeco.javacpp.annotation.Platform;
@@ -765,15 +768,16 @@ public class Loader {
                         File file = new File(directoryOrFile, entryName.substring(jarEntryName.length()));
                         if (entry.isDirectory()) {
                             file.mkdirs();
+                            file.setLastModified(entryTimestamp);
                         } else if (!cacheDirectory || !file.exists() || file.length() != entrySize
-                                || file.lastModified() != entryTimestamp || !file.equals(file.getCanonicalFile())) {
+                                || file.lastModified() != entryTimestamp) {
                             // ... extract it from our resources ...
                             file.delete();
                             String s = resourceURL.toString();
                             URL u = new URL(s.substring(0, s.indexOf("!/") + 2) + entryName);
                             file = extractResource(u, file, prefix, suffix);
+                            file.setLastModified(entryTimestamp);
                         }
-                        file.setLastModified(entryTimestamp);
                     }
                 }
                 return directoryOrFile;
@@ -1247,7 +1251,7 @@ public class Loader {
                     foundLibraries.put(preload, urls = findLibrary(cls, p, preload, pathsFirst));
                 }
                 String filename = null;
-                if (oldUrls == null || urls.length > 0) {
+                if (oldUrls == null && urls.length > 0) {
                     filename = loadLibrary(cls, urls, preload, preloaded.toArray(new String[preloaded.size()]));
                 }
                 if (filename != null && new File(filename).exists()) {
@@ -1341,7 +1345,7 @@ public class Loader {
                     foundLibraries.put(library, urls = findLibrary(cls, p, library, pathsFirst));
                 }
                 String filename = null;
-                if (oldUrls == null || urls.length > 0) {
+                if (oldUrls == null && urls.length > 0) {
                     filename = loadLibrary(cls, urls, library, preloaded.toArray(new String[preloaded.size()]));
                 }
                 if (cacheDir != null && filename != null && filename.startsWith(cacheDir)) {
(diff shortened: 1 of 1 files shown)


src/main/java/org/bytedeco/javacpp/Loader.java
```diff
@@ -765,15 +768,16 @@ public static File extractResource(URL resourceURL, File directoryOrFile,
                         File file = new File(directoryOrFile, entryName.substring(jarEntryName.length()));
                         if (entry.isDirectory()) {
                             file.mkdirs();
+                            file.setLastModified(entryTimestamp);
                         } else if (!cacheDirectory || !file.exists() || file.length() != entrySize
-                                || file.lastModified() != entryTimestamp || !file.equals(file.getCanonicalFile())) {
```
HannesWell, line 769: I'm not sure what the intention of the comparison with the canonical file is, but the file is not canonicalized in the called extractResource() method. Furthermore canonicalization does not (or at least should not) change the effective target the file is pointing to, so the metadata checked before should not change because of that.

src/main/java/org/bytedeco/javacpp/Loader.java
```diff
@@ -1247,7 +1251,7 @@ public static String load(Class cls, Properties properties, boolean pathsFirst,
                     foundLibraries.put(preload, urls = findLibrary(cls, p, preload, pathsFirst));
                 }
                 String filename = null;
-                if (oldUrls == null || urls.length > 0) {
+                if (oldUrls == null && urls.length > 0) {
```
HannesWell, line 1254: I'm not sure the or was intentional or not, but before it called the following loadLibrary method when
1. `oldURLs==null and urls.length>0` , so the library in question was not searched before and found now.
2. `oldURLs==null and urls.length==0`, e.g. the library in question was not searched before but not found.
2. `oldURLs!=null and urls.length>0`, e.g. the library in question was searched before but and found.

For me it does not make sense to attempt to load the library when it was not found (case 2) or was already loaded (case 3). Or did I oversee something?

HannesWell:
Maybe this also useful regarding issues #452.
The most significant improvements from this PR were achieved by avoiding calls to File methods.

src/main/java/org/bytedeco/javacpp/Loader.java
saudet: Symbolic links get created and we need to resolve them to prevent creating cycles.

src/main/java/org/bytedeco/javacpp/Loader.java
saudet: If a previously found library failed to load, we should try to keep finding other versions that might be able to load.

saudet:
> Maybe this also useful regarding issues #452.
> The most significant improvements from this PR were achieved by avoiding calls to File methods.

If you know of a more efficient way without changing the logic, please try to do that! Thanks

src/main/java/org/bytedeco/javacpp/Loader.java
HannesWell: Can you please explain how this prevents cycles or is there an example anywhere? Because I don't understand it.
I'm asking because avoiding the canonicalization had a significant impact on the runtime improvement (around one third of the runtime reduction).
So if we can avoid this call or at least have a cheaper pre-check (e.g. `canCreateSymbolicLink==true`? or `Files.isSymbolicLink()==true`, I'm not sure if this is much faster), it should make it faster.
But of course the logic has to be correct in any case.

saudet: There can be symbolic links anywhere in the path, not just the last file in the path.

I'm pretty sure we can disable those checks when org.bytedeco.javacpp.cachedir.nosubdir is set though, and I think that's what you're already using, right? If so, I'd be happy to skip all that when that flag is set. How does that sound?

saudet:
So, is there anything in particular that is preventing you from making progress with those pull requests?

HannesWell:
> So, is there anything in particular that is preventing you from making progress with those pull requests?

It is just my limited time and shifted priorities, but I definitely plan to complete this PR (as well as my other open PRs!)
I hope/expect to have some free-time around the christmas/new-year days to complete this and the others.

saudet:
BTW, concerning symbolic links on Windows, we could definitively disable them and assume by default that they are never supported. The user could still enable them with some system property, but since they are not available to users by default, and with bugs such as https://bugs.openjdk.java.net/browse/JDK-8003887 that seem like they are never going to get fixed but that @devjeonghwan found can still cause problems, it doesn't sound like we should be relying on them for anything on Windows. But we should still be able to enable their use in JavaCPP for users that really need them.

saudet:
Based on suggestions in this pull request, I've updated a couple of things in commit https://github.com/bytedeco/javacpp/commit/e8b5734f7d0469478f399bb58e0fd3cfe4af1906.  With these modifications, it takes a bit less than 1200 ms on my Windows 10 machine to execute the following, when everything is already cached, extracted, and all:
```java
        Py_Initialize(org.bytedeco.scipy.presets.scipy.cachePackages());
        org.bytedeco.numpy.global.numpy._import_array();
```
I haven't tried to do something about the unnecessary calls to loadLibrary(), but it wouldn't even gain us 100 ms, so I'm reluctant to change the logic there just for that. If you see other places where we can gain more, please update this pull request! Thanks

@yukoba Does your javacpp-embedded-python incur any additional overhead that we should know about?

HannesWell:
Thanks for implementing the changes and sorry for the delay. Seem that my estimation was bad but I hope to be able to check for possible updates of this and my other PR/issues in the next few weeks.

PR #519, opened 2021-10-15
Fix symbolic file(or contains symbolic directory) check condition

devjeonghwan:
`file.getCanonicalFile()` is not solve symbolic link in Windows environment (See [issue](https://bugs.openjdk.java.net/browse/JDK-8003887))

So i change to more precise method `toRealPath()`.


saudet:
Let's replace all the calls, not just that one. However, `Path.toRealPath()` doesn't appear available before Android 8.0:
https://developer.android.com/reference/java/nio/file/Path?hl=ro#toRealPath(java.nio.file.LinkOption...)
So let's create our own `Loader.getCanonicalFile()` that calls `Path.toRealPath()` only on Windows, and `File.getCanonicalFile()` on other platforms.

saudet:
No, let's just leave it as a string. That just adds unnecessary complexity.

PR #552, opened 2022-02-08
First attempt for auto generated downcast code based on c++ type_index

nyholku:
I've modified the code so that we don't touch the Parser, everything is done in Generator, only one annotation that can/needs to be attached to the handle or pointer to the class hierarchy base(s). I doubt some of the logic I used is 100% full proof but we gotta start from somewhere.


nyholku:
Hiw annoying, despite my best efforts and very nice diff on my local repo the pull request diff seem brain dead. I'll work on this later.

Issue #557, opened 2022-02-24
Pointer.asByteBuffer() behavior on large arrays.


masterav10:
Discovered this while attempting to use a large BytePointer as a native cache. When attempting to get a ByteBuffer from the higher end of the array (position > Integer.MAX_VALUE), the buffer's capacity, limit, and position get truncated or is otherwise invalid.

A simple test case to demonstrate the problem:

```
    @Test public void testByteBufferIndexingUsingLargePointers() {
        final long bigSize = Integer.MAX_VALUE + 1L;
        
        try(BytePointer pointer = new BytePointer(bigSize))
        {
            pointer.fill(1);
            pointer.put(pointer.capacity() - 1L, (byte) 100);
            
            ByteBuffer buffer = pointer.position(Integer.MAX_VALUE)
                    .asByteBuffer();
            
            assertEquals(0L, buffer.position());
            assertEquals(1L, buffer.remaining());
            assertEquals((byte)100, buffer.get());
        }
    }
```
An easy solution would be to reimplement the method to work like ByteBuffer.slice(). The returned ByteBuffer would have a position of 0, limit + capacity of remaining(), and we simply shift the address by the appropriate number of bytes.

My assessment is that fixing this will have a relatively low impact on client code; it usually doesn't make sense to mess with position() and limit() in absolute numbers if at all. I imagine most people are using the returned ByteBuffer as a way to read the data stored in the pointer.

Unless there is something glaring I'm going to take stab at fixing this.

saudet:
The thing with Pointer.asByteBuffer() is that it's meant to copy the address and position as is...

Anyway, since Pointer.getPointer() has been introduced, I don't recommend using Pointer.position() for anything else than advanced optimization. So, just use Pointer.getPointer(), it should already do what you need it to do here, but please let me know if this isn't the case.

masterav10:
Your presented solution does indeed work.

It's still a bug based on how the method is documented. I don't like the idea of throwing an exception if one of the parameters is outside the supported range of a ByteBuffer.

> The thing with Pointer.asByteBuffer() is that it's meant to copy the address and position as is...

Is there some benefit of preserving the address and position? Other than for posterity.

saudet:
Right, the docs could be updated to clarify more explicitly what it's doing. Mind sending a pull request for that?

As for not changing the behavior, well, for backward compatibility. There's no reason to "fix" this because it works with Pointer.getPointer(), which we should be using anyway.

[...]

PR #558, opened 2022-02-24
Improving the documentation of Pointer.asByteBuffer()

masterav10:
Provide examples of how it can be misused or used properly.

This pull request was made to document the issue and workaround discovered in #557.

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/Pointer.java
+++ b/src/main/java/org/bytedeco/javacpp/Pointer.java
@@ -859,11 +859,63 @@ public class Pointer implements AutoCloseable {
 
     private native ByteBuffer asDirectBuffer();
     /**
-     * Returns a ByteBuffer covering the memory space from this.position to this.limit.
-     * If limit == 0, it uses position + 1 instead. The way the methods were designed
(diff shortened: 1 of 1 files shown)


src/main/java/org/bytedeco/javacpp/Pointer.java
```diff
@@ -859,11 +859,61 @@ public int sizeof() {
 
     private native ByteBuffer asDirectBuffer();
     /**
-     * Returns a ByteBuffer covering the memory space from this.position to this.limit.
-     * If limit == 0, it uses position + 1 instead. The way the methods were designed
-     * allows constructs such as {@code this.position(13).limit(42).asByteBuffer()}.
```
saudet, line 864: Let's not remove the correct information that is already there. Please only add new information.

masterav10: Did you want that precise verbiage? I updated some pieces to match the documentation found in the ByteBuffer classes.

saudet: No, your update looks fine! Thanks

Issue #236, opened 2018-05-02
Attach as Daemon


masterav10:
I'd like for the JNI code to attach as a daemon thread. This will allow my applications to cleanly exit when using the hack provided #143. It will also be useful when I finish my bindings for Decklink, which will likely be function pointer based. Does anyone foresee an issue with doing this? I'd be happy to put up a pull request and test it with my build of FFmpeg later this week.

[See this line in the generator.](https://github.com/bytedeco/javacpp/blob/96045c9b86625df009e3986c555e467bd7d58853/src/main/java/org/bytedeco/javacpp/tools/Generator.java#L1160)

saudet:
The only thing that makes a deamon thread different from a normal thread is that the JVM will terminate the process even when daemon threads are still running. If that's not an issue for your application or your library, there's nothing special about JavaCPP here. Are you worried about anything in particular?

masterav10:
Wasn't sure if exiting while during a callback transition had the potential to cause a JVM crash. I'm guessing this is unlikely to be a problem, as that is the whole point of the daemon. I'll make the change.

saudet:
Some native libraries might not like it and crash the JVM. That's entirely
possible.

saudet:
So, in the end, have you encountered any issue?

masterav10:
I made the change and tested it against my Decklink FFmpeg build. When I go to register the logging callback, it seems to block indefinitely. There are still a couple of things I can try, however. I'll keep you posted.

saudet:
What does the code look like?

PR #561, opened 2022-03-07
Attach as Daemon Thread

masterav10:
Introduces a new #define JNI_ATTACH_AS_DAEMON, which will attach threads to java as daemons. This will generally be used in combination with NO_JNI_DETACH_THREAD to ensure that lingering threads do no hold up the parent from exiting.

```
@Platform(
    define={
         "NO_JNI_DETACH_THREAD", "JNI_ATTACH_AS_DAEMON"
    }
```

Primarily intended to be used with long lived callbacks that get called at a high frequency. Personally going to use this with the Decklink bindings I'm preparing.

See #236 for additional details.


saudet:
Looks good, thanks! Are there any reasons we shouldn't just make this the default?

BTW, NO_JNI_DETACH_THREAD isn't needed on Linux and Mac anymore. They do not get detached before they exit anyway.

masterav10:
> Are there any reasons we shouldn't just make this the default?

I suppose it is possible for the an application to terminate mid-callback when the thread is a daemon. This is not possible with the default behavior, since the thread will only disappear once the callback is complete. I don't know if this an actual problem though.

I'm fine either way, let me know what you prefer and it will be done.

> BTW, NO_JNI_DETACH_THREAD isn't needed on Linux and Mac anymore. They do not get detached before they exit anyway.

So threads are reused by default on these platforms? Won't that prevent the application from closing normally (forcing use of System.exit()).

saudet:
> So threads are reused by default on these platforms? Won't that prevent the application from closing normally (forcing use of System.exit()).

Maybe that's why we shouldn't try to consider those threads to be managed by Java and initialize them as daemon...

masterav10:
Let me make some tests to see if there are any issues.

saudet:
I think the whole concept of "daemon threads" in Java is to replace the implicit call to exit() when returning from the main() function in C. Java doesn't do that, so they had to come up with some other convention, and that's what they ended up settling on I guess. In a native application, all threads are therefore in that sense "daemon" since whenever we return from the main() function, the program exits. Consequently, I think it makes sense to consider all those threads created on the native side to be "daemon", since native libraries expect them to work that way anyway.

masterav10:
I'm pretty happy with the changes, assuming they build. Threads are now always attached as daemons. See the associated test case; I've independently verified using my Decklink Bindings.

Side note; it might be worth tracking thread attachment using thread_local storage. My understanding is that thread_local is automatically destroyed upon thread termination; meaning we can detach in the destructor. If this works, then we can make NO_JNI_DETACH_THREAD the default behavior for windows as well. 

Thoughts? Should be pretty easy to implement and test; I don't mind doing it here or in another branch.

saudet:
> Side note; it might be worth tracking thread attachment using thread_local storage. My understanding is that thread_local is automatically destroyed upon thread termination; meaning we can detach in the destructor. If this works, then we can make NO_JNI_DETACH_THREAD the default behavior for windows as well.
> 
> Thoughts? Should be pretty easy to implement and test; I don't mind doing it here or in another branch.

Sounds like a good idea! If it works, we can try to make this the default on all platforms when C++11 is detected.

saudet:
BTW, even if you implement detach using thread_local, you may very well end up using NO_JNI_DETACH_THREAD in your application anyway to avoid overhead from that. So let's merge this and you can work on it when you have time?

masterav10:
I can do it separately.

PR #562, opened 2022-03-13
Automatic Thread Detachment for Windows

masterav10:
New Default Behavior: vm->DetachCurrentThread() is called automatically when an attached native thread is terminated. Previously, a new thread was attached and detached per method call, creating memory pressure and increasing debugging complexity.

New behavior attaches the thread once, and relies on a static _declspec( thread ) object to detect thread termination.

NO_JNI_DETACH_THREAD retains its original behavior, although it is likely unnecessary.


src/main/java/org/bytedeco/javacpp/tools/Generator.java
```diff
@@ -1460,6 +1472,9 @@ boolean classes(boolean handleExceptions, boolean defineAdapters, boolean conver
             out.println("            *env = NULL;");
             out.println("            goto done;");
             out.println("        }");
+            out.println("#if !defined(NO_JNI_DETACH_THREAD) && defined(_WIN32)");
+            out.println("       windowsJniAutoDetachInstance.attached = true;");
```
saudet, line 1476: Let's store here the `env` pointer itself instead of just a flag, to be consistent with how it's done with pthreads.

saudet: ... along with something like this?
```java
            out.println("#if !defined(NO_JNI_DETACH_THREAD) && defined(_WIN32)");
            out.println("    if ((*env = windowsJniAutoDetachInstance.env) != NULL) {");
            out.println("        attached = true;");
            out.println("        goto done;");
            out.println("    }");
            out.println("#endif");
```

masterav10: Whoops, missed that short circuit. Good find.

PR #574, opened 2022-05-27
add getUnsigned() and putUnsigned() methods.

before31:
add getUnsigned() and putUnsigned() methods to get the correct unsigned integer value directly.


saudet:
Maybe we also want to add getUnsignedShort(), getUnsignedInt() etc while we're at it? Or are you OK with just bytes for now?

before31:
When you say `getUnsignedInt()`, you mean:
 - I need to get the next **FOUR** bytes and combine them into an integer? or, 
 - I get the next **ONE** byte and operate '& 0xFF' to convert it to an unsigned integer(0-255)?

In my `getUnsigned()` implementation, it is the later one. The `getUnsigned` method actually returns a integer value in the range 0-255, and my need is that.

saudet:
Your getUnsigned() is fine, I'm talking about 16, 32, and 64-bit values. You're never going to need those?

before31:
So far, I may not need those.

Issue #609, opened 2022-09-27
building with clang 15 on OSX


jkhoogland:
Hi,
I use javacpp to build a java wrapper around a c++ library. 
With the standard apple clang everything builds fine.

When I tried building using clang 15.0 on OSX, it fails out with the following message:

```
jnilibrary.cpp:66:29: error: 'TARGET_OS_IPHONE' is not defined, evaluates to 0 [-Werror,-Wundef-prefix=TARGET_OS_]
#if defined(__ANDROID__) || TARGET_OS_IPHONE
                            ^
1 error generated.
```

I tried to insert a line in the preset file to define `TARGET_OS_IPHONE` but that did not help.

```
define = {
    "TARGET_OS_IPHONE 0"
    },
```

There is some SO [answer](https://stackoverflow.com/questions/3742525/target-os-iphone-and-applicationtests) which suggests to include 

```
#include <TargetConditionals.h>
```

but I am not sure where I would put that as this seems something deep inside the javacpp magic.

Did anybody hit this problem and find a fix ?

wabscale:
I believe this may be because the jnilibrary.cpp file that is generated has this near the top:

```cpp
#include <jni.h>

#ifdef __ANDROID__
    #include <android/log.h>
#elif defined(__APPLE__) && defined(__OBJC__)
    #include <TargetConditionals.h>
    #include <Foundation/Foundation.h>
#endif
```

My guess is that on MacOS, if you brew install llvm and get the non-apple version of clang the `__OBJC__` is undefined.

saudet:
The fix for this has been released with JavaCPP 1.5.8. Thanks for the contribution!

PR #610, opened 2022-09-27
Fix building with clang 15 on OSX #609

wabscale:
The problem is explained in the issue, but to reiterate: 

Newer versions of apple's clang will error if there is an undefined reference to TARGET_OS_*. The solution to this is to include the `TargetConditionals.h` file ahead of any references. The problem originates here: 

https://github.com/bytedeco/javacpp/blob/f77f8d77f3724128bc26440d930b786892c58312/src/main/java/org/bytedeco/javacpp/tools/Generator.java#L313-L318

The `TargetConditionals.h` is only included if both `__APPLE__` and `__OBJC__` are definied.  If you use a brew installed llvm (currently version 15), the `__APPLE__` will be defined but `__OBJC__` is not. 

What I have done is made it so that `TargetConditionals.h` is always included on apple, and the `Foundation/Foundation.h` is only included when `__OBJC__` is defined. The result is this:

```cpp
#ifdef __ANDROID__
    #include <android/log.h>
#elif defined(__APPLE__)
    #include <TargetConditionals.h>
#endif

#if defined(__OBJC__) 
    #include <Foundation/Foundation.h>
#endif
```

I've tested this on OSX 12.6, clang 15 (installed via `brew install llvm`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 
+ * Fix `Generator` compiler errors on Mac for Clang without Objective-C support ([pull #610](https://github.com/bytedeco/javacpp/pull/610))
  * Prevent `Parser` from outputting cast methods for base classes that are `Info.skip` ([pull #607](https://github.com/bytedeco/javacpp/pull/607))
  * Ensure `Generator` and `Parser` process header files from `cinclude` before `include` ([issue #580](https://github.com/bytedeco/javacpp/issues/580))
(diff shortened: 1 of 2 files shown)


saudet:
Thanks for the fix! To make sure we don't break anything, we probably want more something like this though:
```patch
@@ -315,6 +315,8 @@ public class Generator {
         out.println("#elif defined(__APPLE__) && defined(__OBJC__)");
         out.println("    #include <TargetConditionals.h>");
         out.println("    #include <Foundation/Foundation.h>");
+        out.println("#elif defined(__APPLE__)");
+        out.println("    #include <TargetConditionals.h>");
         out.println("#endif");
         out.println();
         out.println("#ifdef __linux__");
```

wabscale:
Great, I've overwritten the commit with what you have requested.

PR #612, opened 2022-10-06
Problem: nvcc does not respect __declspec(thread)

masterav10:
Fixes issue described here: https://github.com/bytedeco/javacpp-presets/issues/1239

Uses thread_local if available, which seems to behave better when nvcc is compiling.


saudet:
It sounds like NVCC on Windows is now using Clang as backend instead of MSVC. That would explain a couple of things since their version of `__declspec(thread)` doesn't support destructors: https://clang.llvm.org/docs/AttributeReference.html#thread 

So, instead of checking for C++11, let's check for CUDA on Windows, with something like 
```
#if !defined(NO_JNI_DETACH_THREAD) && defined(_WIN32) && defined(__CUDACC__)
...
#elif !defined(NO_JNI_DETACH_THREAD) && defined(_WIN32)
...
```

masterav10:
I am certain this not the case, at least on my machine. If I run from plain old command line, I get the following:

` nvcc fatal   : Cannot find compiler 'cl.exe' in PATH`

The documentation [here](https://docs.nvidia.com/cuda/archive/11.6.2/cuda-installation-guide-microsoft-windows/index.html) makes not mention of clang either.

saudet:
Yes, it does: https://developer.nvidia.com/cuda-llvm-compiler

It probably gets preprocessed by clang, at the very least.

saudet:
Actually, no, it doesn't mention Clang. It looks like it gets preprocessed by cudafe:
https://arcb.csc.ncsu.edu/~mueller/cluster/nvidia/2.0/nvcc_2.0.pdf
And the string "thread-local variable cannot be dynamically initialized" is found in cudafe++.exe, so that sounds like a bug in there, and I think adding a special case for it is the right thing to do.

masterav10:
I think we are in business now. Let me know if you'd like anything else changed.

saudet:
It looks like we can use `thread_local` and `__declspec(thread)` interchangeably in the same positions, so we can simplify this a bit.

masterav10:
And so we can; see the new changes.

saudet:
Yeah, but we only really care about that on Windows, and I've also noticed that GCC doesn't support `__declspec(thread)`, so we might as well use `thread_local` when we have a C++11 compiler, and only fall back on `__declspec(thread)`. What about something like this?
```patch
@@ -525,7 +525,11 @@ public class Generator {
         out.println("}");
         out.println();
         out.println("#if !defined(NO_JNI_DETACH_THREAD) && defined(_WIN32)");
+        out.println("#if __cplusplus >= 201103L || _MSC_VER >= 1900");
+        out.println("   static thread_local struct JavaCPP_thread_local {");
+        out.println("#else");
         out.println("   static __declspec(thread) struct JavaCPP_thread_local {");
+        out.println("#endif");
         out.println("       JNIEnv* env = NULL;");
         out.println("       ~JavaCPP_thread_local() {");
         out.println("           if (env && JavaCPP_vm) {");
```

src/main/java/org/bytedeco/javacpp/tools/Generator.java
```diff
@@ -293,6 +294,11 @@ boolean classes(boolean handleExceptions, boolean defineAdapters, boolean conver
         out.println("    typedef int jint;");
         out.println("    typedef long long jlong;");
         out.println("    typedef signed char jbyte;");
+        out.println("    #if __cplusplus >= 201103L || _MSC_VER >= 1900");
+        out.println("        #define THREADLOCAL thread_local");     
+        out.println("    #else");
+        out.println("        #define THREADLOCAL __declspec(thread)");     
+        out.println("    #endif");
```
saudet, line 301: I'd rather not define a new symbol since we're only using it in a single place.

src/main/java/org/bytedeco/javacpp/tools/Generator.java
```diff
@@ -525,14 +531,14 @@ boolean classes(boolean handleExceptions, boolean defineAdapters, boolean conver
...
-        out.println("   static __declspec(thread) struct JavaCPP_thread_local {");
+        out.println("   static struct JavaCPP_thread_local {");
         out.println("       JNIEnv* env = NULL;");
         out.println("       ~JavaCPP_thread_local() {");
         out.println("           if (env && JavaCPP_vm) {");
         out.println("               JavaCPP_vm->DetachCurrentThread();");
         out.println("           }");
         out.println("       }");
-        out.println("   } JavaCPP_thread_local; ");
+        out.println("   } THREADLOCAL JavaCPP_thread_local; ");
```
saudet, line 541: It looks like it's more standard to use that keyword before the type, for example:
https://en.cppreference.com/w/cpp/language/storage_duration
It's like a modifier of the `static` keyword.

src/main/java/org/bytedeco/javacpp/tools/Generator.java
masterav10: I will change it if you like your way better. Personally, I find the nested macro difficult to read.

saudet: This is also a nested macro, isn't? We can make the other one a bit better, for example, I think this works:

```patch
@@ -525,7 +525,12 @@ public class Generator {
         out.println("}");
         out.println();
         out.println("#if !defined(NO_JNI_DETACH_THREAD) && defined(_WIN32)");
-        out.println("   static __declspec(thread) struct JavaCPP_thread_local {");
+        out.println("#if __cplusplus >= 201103L || _MSC_VER >= 1900");
+        out.println("   thread_local");
+        out.println("#else");
+        out.println("   __declspec(thread)");
+        out.println("#endif");
+        out.println("   static struct JavaCPP_thread_local {");
         out.println("       JNIEnv* env = NULL;");
         out.println("       ~JavaCPP_thread_local() {");
         out.println("           if (env && JavaCPP_vm) {");
```

masterav10: Does not compile. Looks like it needs to come before static.

saudet: Ok, so this works for sure and doesn't look bad:
```patch
@@ -525,7 +525,12 @@ public class Generator {
         out.println("}");
         out.println();
         out.println("#if !defined(NO_JNI_DETACH_THREAD) && defined(_WIN32)");
-        out.println("   static __declspec(thread) struct JavaCPP_thread_local {");
+        out.println("#if __cplusplus >= 201103L || _MSC_VER >= 1900");
+        out.println("   static thread_local");
+        out.println("#else");
+        out.println("   static __declspec(thread)");
+        out.println("#endif");
+        out.println("   struct JavaCPP_thread_local {");
         out.println("       JNIEnv* env = NULL;");
         out.println("       ~JavaCPP_thread_local() {");
         out.println("           if (env && JavaCPP_vm) {");
```

Issue #628, opened 2022-10-24
Tool to get the address of a DirectByteBuffer


calvertdw:
I don't know if it is safe or a good idea, but I'm often wanting to get the address of a direct ByteBuffer that I already have in order to setup a JavaCPP Pointer to point there. Currently, we are able to do this by creating a `new Pointer(ByteBuffer)`, however, in realtime and high performance settings, we don't want to allocate a new Pointer each time.

I think the way it's done internally is using `env->GetDirectBufferAddress`.
https://github.com/bytedeco/javacpp/blob/81252dd2822d2bee3555bb4c094a78e94796b4db/src/main/java/org/bytedeco/javacpp/tools/Generator.java#L2333

Maybe we could have some static method available that just allows us to utilize `GetDirectBufferAddress`?

saudet:
You could easily add a mapping for GetDirectBufferAddress() in your library, after all that's JavaCPP is for :)

But if we were to add something to the Pointer class, what about a `protected static native long Pointer.getDirectBufferAddress(Buffer b)`? Could you open a pull request to that effect?

calvertdw:
It seems generally valuable for each preset to have available and in most places we are using the provided presets anyway. I like the idea of a static method in Pointer.

I'm going to see if @ds58 has time for this. He's been helping our lab use JavaCPP to it's full potential. We've wrapped a few custom things already and it's been a huge success.

PR #629, opened 2022-10-26
Add static access to JNI GetDirectBufferAddress

ds58:
Addresses #628

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/Pointer.java
+++ b/src/main/java/org/bytedeco/javacpp/Pointer.java
@@ -33,6 +33,7 @@ import java.nio.ByteOrder;
 import java.util.Iterator;
 import java.util.concurrent.atomic.AtomicInteger;
 import org.bytedeco.javacpp.annotation.Name;
+import org.bytedeco.javacpp.annotation.Raw;
 import org.bytedeco.javacpp.tools.Generator;
 import org.bytedeco.javacpp.tools.Logger;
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/bytedeco/javacpp/Pointer.java
```diff
@@ -589,6 +590,10 @@ public static long maxPhysicalBytes() {
     /** Returns the amount of physical memory that is free according to the operating system, or 0 if unknown. */
     @Name("JavaCPP_availablePhysicalBytes") public static native long availablePhysicalBytes();
 
+    /** Returns the starting address of the memory region referenced by the given direct {@link java.nio.Buffer}.
+     * An alternative to allocating a new Pointer object if all you need is the memory address. */
+    @Name("JavaCPP_getDirectBufferAddress") public static native long getDirectBufferAddress(@Raw(withEnv = true) Buffer b);
```
saudet, line 595: Can we make this protected? It doesn't feel like something that should be used outside of a wrapper.

calvertdw: By wrapper, do mean a class that extends Pointer? I suppose that's a good point, since we're just using this to set the address of our MutableBytePointer class, which allows publicly setting the address. We can just expose a method in that one that uses this internally.

src/main/java/org/bytedeco/javacpp/tools/Generator.java
```diff
@@ -888,6 +888,10 @@ boolean classes(boolean handleExceptions, boolean defineAdapters, boolean conver
             out.println("    env->DeleteGlobalRef(globalRef);");
             out.println("}");
             out.println();
+            out.println("static inline jlong JavaCPP_getDirectBufferAddress(JNIEnv *env, jclass cls, jobject obj) {");
+            out.println("    return (jlong)env->GetDirectBufferAddress(obj);");
+            out.println("}");
+            out.println();
```
saudet, line 894: Let's put this just after JavaCPP_availablePhysicalBytes() for consistency.

PR #635, opened 2022-11-23
Thread Local Attach/Detach

masterav10:
In 1.5.8, we introduced thread_local storage and daemon threads for automatically detaching the JVM on native termination. We then proceeded to wrap those calls in NO_JNI_DETACH_THREAD, which completely defeated the implementation.

This pull requests removes the use of NO_JNI_DETACH_THREAD from any thread_local or equivalent objects, since there is no reason why you would want to keep the attached thread around.

Also the behavior of the thread_local destructors matches the code used in JavaCPP_detach(...)

Putting this up now for review; we may need to make an additional test case to verify the behavior but I'm not sure.


saudet:
NO_JNI_DETACH_THREAD was already there before, that's fine. Why do you want to remove if?

masterav10:
I had a project that was declaring NO_JNI_DETACH_THREAD for callbacks prior to 1.5.8. A new thread was attaching and detaching to each call, which was severely reducing performance. To work around this, I was using NO_JNI_DETACH_THREAD. This was still suboptimal, since terminating the callback left the thread hanging around, even though the native thread was gone.

If the thread_local or equivalent construct is available, then NO_JNI_DETACH_THREAD is no longer necessary, since the most efficient behavior is implemented.

In the absence of thread_local, NO_JNI_DETACH_THREAD still works in the original way.

This is more a cleanup than anything else.

saudet:
Some users may want to opt out of the overhead, or have an environment that
doesn't support that. It still makes sense to leave that there.

masterav10:
Okay, I will close this then.

PR #649, opened 2023-01-25
Add support for friend functions

HGuillemet:
This PR is a proposal to add support for friend functions to the parser.
Currently, these functions are simply ignored.

Friend functions are non-member functions but are declared in a class body. They have access to the class private and protected members.

This PR does the following:
1) When a friend function is parsed, instead of being added to the current list of declarations, it’s added to a special list of non-member declarations.
2) When we finish parsing an entire group, and if the declaration of the group has been successfully added to the parent declaration list, we also add the non-members declarations to the global declaration list.

Tested on Pytorch : this patch allows to map about 25 functions that were missing, and doesn’t alter anything else. Most of these functions are operator overloads. The generated new JNI code compiles, baring the removal of [this info](https://github.com/bytedeco/javacpp-presets/blob/a48eb57460d452825473410fd3cc3c870716203f/pytorch/src/main/java/org/bytedeco/pytorch/presets/torch.java#L2076). It corresponds to a template instantiation apparently not meant to exist. Any idea why it was added ? The second infoMap `c10::Dict<c10::Ivalue,c10::Ivalue>::iterator` is correct.

One minor caveat:
C++ allows operators to be declared either as friend functions or as member functions, e.g.:
```c++
struct Foo {
    private:
      int n;
    public:
      Foo(int x) {
        n=x;
      }
      bool operator<=(Foo& rhs) {
        return n <= rhs.n;
      }
   friend bool operator<(Foo& lhs,  Foo& rhs) {
     return lhs.n < rhs.n
   }
};
```
In both case, the C++ user will write the same operation : `x < y` or `x ≤ y` but in Java it will map as an instance method `x.lessThanEquals(y)` and as a static method `lessThan(x, y)`. Not a big deal.

Another caveat is about accessibility: C++ has some complicated rules for locating a function in the tree of namespaces based on the type of the arguments: the ADL (argument-dependent lookup).
```c++
namespace ns {
  struct Bar {
    …
  };
  struct Foo {
    friend bool operator<(Foo& lhs, Foo& rhs) { … }
    friend bool operator<(Bar& lhs, Bar& rhs) { … }
  };
}
```
in the exampe above,  `x < y ` works if x and y are Foo, but not it they are Bar, because the declaration of < on Bars cannot be found from type Bar. Ideally the first operator must be mapped to Java, because it is accessible, but not the second one.
In this PR we map the function only if it has an argument of the same type that the one it’s declared in. That covers only a subset of the ADL and some theoretically accessible friend functions could be missed.

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/tools/Parser.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Parser.java
@@ -2332,8 +2332,16 @@ public class Parser {
                 break;
             }
         }
-        if (tokens.get().match("&&") || (context.javaName == null && localNamespace > 0) || (info != null && info.skip)) {
-            // this is a rvalue function, or a member function definition or specialization, skip over
+        Info info2 = infoMap.getFirst(null);
+        boolean mapFriends = info != null ? info.mapFriends : info2 != null ? info2.mapFriends : false;
+        if (tokens.get().match("&&")
+                || (context.javaName == null && localNamespace > 0)
+                || (info != null && info.skip)
+                || (type.friend && !mapFriends)
+        ) {
+            // this is a rvalue function, or a member function definition or specialization,
+            // or a friend function and we don't want to map it,
+            // skip over
             while (!tokens.get().match(':', '{', ';', Token.EOF)) {
                 tokens.next();
             }
--- a/src/main/java/org/bytedeco/javacpp/tools/Parser.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Parser.java
@@ -2114,9 +2114,7 @@ public class Parser {
                     params.list += "/*" + defaultToken + defaultValue + "*/";
                 }
                 params.signature += '_';
-                for (char c : dcl.type.javaName.substring(dcl.type.javaName.lastIndexOf(' ') + 1).toCharArray()) {
-                    params.signature += Character.isJavaIdentifierPart(c) ? c : '_';
-                }
+                params.signature += dcl.type.signature();
                 params.names += (count > 1 ? ", " : "") + paramName;
                 if (dcl.javaName.startsWith("arg")) {
                     try {
@@ -2394,6 +2392,7 @@ public class Parser {
         boolean first = true;
         for (int n = -2; n < Integer.MAX_VALUE; n++) {
             decl = new Declaration();
+            Declaration extraDecl = null; // Secondary declaration to add. Currently only used for friends.
             tokens.index = startIndex;
             boolean useDefaults = (info == null || !info.skipDefaults) && n % 2 != 0;
             if ((type.constructor || type.destructor || type.operator) && params != null) {
@@ -2524,6 +2523,7 @@ public class Parser {
                If it does not, skip over. */
             if (type.friend && friendly) {
                 Declarator[] paramDeclarators = dcl.parameters.declarators;
+                String signature = dcl.javaName;
                 String argList = "", staticArgList = "";
                 boolean foundThis = false;
                 for (Declarator paramDecl : paramDeclarators) {
@@ -2538,11 +2538,15 @@ public class Parser {
                             argList += ", ";
                         }
                         argList += paramDecl.type.javaName + " " + paramDecl.javaName;
+                        signature += '_' + paramDecl.type.signature();
                         staticArgList += paramDecl.javaName;
                     }
                 }
                 if (foundThis) {
-                    decl.text += "public " + dcl.type.javaName + " " + dcl.javaName + "(" + argList + ") { "
+                    extraDecl = new Declaration();
+                    extraDecl.signature = signature;
+                    extraDecl.declarator = dcl; // Used in group to recognize friends
+                    extraDecl.text = "public " + dcl.type.javaName + " " + dcl.javaName + "(" + argList + ") { "
                               + (dcl.type.javaName.equals("void") ? "" : "return ") + dcl.javaName + "(" + staticArgList + "); }\n";
                 } else {
                     friendly = false;
@@ -2619,6 +2623,7 @@ public class Parser {
             if (dcl.javaName.length() > 0 && !found && (!type.destructor || (info != null && info.javaText != null))) {
                 if (declList.add(decl, fullname)) {
                     first = false;
+                    if (extraDecl != null) declList.add(extraDecl);
                 }
                 if (type.virtual && context.virtualize) {
                     break;
@@ -3651,7 +3656,8 @@ public class Parser {
             }
         }
         for (Declaration d : declList2) {
-            if (!d.inaccessible && (d.declarator == null || d.declarator.type == null
+            if ((!d.inaccessible || d.declarator != null && d.declarator.type.friend)
+                    && (d.declarator == null || d.declarator.type == null
                     || !d.declarator.type.constructor || !abstractClass || (info != null && info.virtualize))) {
                 decl.text += d.text;
             }
--- a/src/main/java/org/bytedeco/javacpp/tools/Parser.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Parser.java
@@ -3670,11 +3670,11 @@ public class Parser {
             info.javaText = decl.text;
         }
         if (declList.add(decl)) {
-            for (Declaration dd: declList2.nonMemberDeclarations) {
+            for (Declaration friendDecl: declList2.nonMemberDeclarations) {
                 // Only add the friend functions that we know will be visible through ADL.
-                for (int i=0; i<dd.declarator.parameters.declarators.length; i++)
-                    if (dd.declarator.parameters.declarators[i].type.cppName.equals(type.cppName)) {
-                        globalDeclList.add(dd);
+                for (Declarator paramDecl: friendDecl.declarator.parameters.declarators)
+                    if (paramDecl.type.cppName.equals(type.cppName)) {
+                        globalDeclList.add(friendDecl);
                         break;
                     }
             }
--- a/src/main/java/org/bytedeco/javacpp/tools/Parser.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Parser.java
@@ -93,6 +93,7 @@ public class Parser {
     InfoMap leafInfoMap = null;
     TokenIndexer tokens = null;
     String lineSeparator = null;
+    DeclarationList globalDeclList;
 
     String translate(String text) {
         Info info = infoMap.getFirst(text);
@@ -2331,8 +2332,8 @@ public class Parser {
                 break;
             }
         }
-        if (type.friend || tokens.get().match("&&") || (context.javaName == null && localNamespace > 0) || (info != null && info.skip)) {
-            // this is a friend declaration, an rvalue function, or a member function definition or specialization, skip over
+        if (tokens.get().match("&&") || (context.javaName == null && localNamespace > 0) || (info != null && info.skip)) {
+            // this is a rvalue function, or a member function definition or specialization, skip over
             while (!tokens.get().match(':', '{', ';', Token.EOF)) {
                 tokens.next();
             }
@@ -2376,7 +2377,7 @@ public class Parser {
             decl.function = true;
             declList.add(decl);
             return true;
-        } else if (type.staticMember || context.javaName == null) {
+        } else if (type.staticMember || type.friend || context.javaName == null) {
             modifiers = "public " + ((info != null && info.objectify) || context.objectify ? "" : "static ") + "native ";
             if (tokens.isCFile) {
                 modifiers = "@NoException " + modifiers;
@@ -2580,7 +2581,7 @@ public class Parser {
                 found |= dcl.signature.equals(d.signature);
             }
             if (dcl.javaName.length() > 0 && !found && (!type.destructor || (info != null && info.javaText != null))) {
-                if (declList.add(decl, fullname)) {
+                if (declList.add(decl, fullname, type.friend)) {
                     first = false;
                 }
                 if (type.virtual && context.virtualize) {
@@ -3660,7 +3661,16 @@ public class Parser {
         } else if (info != null && info.flatten) {
             info.javaText = decl.text;
         }
-        declList.add(decl);
+        if (declList.add(decl)) {
+            for (Declaration dd: declList2.nonMemberDeclarations) {
+                // Only add the friend functions that we know will be visible through ADL.
+                for (int i=0; i<dd.declarator.parameters.declarators.length; i++)
+                    if (dd.declarator.parameters.declarators[i].type.cppName.equals(type.cppName)) {
+                        globalDeclList.add(dd);
+                        break;
+                    }
+            }
+        }
         return true;
     }
 
@@ -4330,12 +4340,12 @@ public class Parser {
 
         String[] includePaths = paths.toArray(new String[paths.size() + includePath.length]);
         System.arraycopy(includePath, 0, includePaths, paths.size(), includePath.length);
-        DeclarationList declList = new DeclarationList();
+        globalDeclList = new DeclarationList();
         for (String include : allIncludes) {
             if (!clsIncludes.contains(include)) {
                 boolean isCFile = cIncludes.contains(include);
                 try {
-                    parse(context, declList, includePaths, include, isCFile);
+                    parse(context, globalDeclList, includePaths, include, isCFile);
                 } catch (FileNotFoundException e) {
                     if (excludes.contains(include)) {
                         // don't worry about missing files found in "exclude"
@@ -4346,14 +4356,14 @@ public class Parser {
                 }
             }
         }
-        declList = new DeclarationList(declList);
+        globalDeclList = new DeclarationList(globalDeclList);
         if (clsIncludes.size() > 0) {
-            containers(context, declList);
+            containers(context, globalDeclList);
             for (String include : clsIncludes) {
                 if (allIncludes.contains(include)) {
                     boolean isCFile = cIncludes.contains(include);
                     try {
-                        parse(context, declList, includePaths, include, isCFile);
+                        parse(context, globalDeclList, includePaths, include, isCFile);
                     } catch (FileNotFoundException e) {
                         if (excludes.contains(include)) {
                             // don't worry about missing files found in "exclude"
@@ -4366,7 +4376,7 @@ public class Parser {
             }
         }
 
-        if (declList.size() == 0) {
+        if (globalDeclList.size() == 0) {
             logger.info("Nothing targeted for " + globalFile);
             return null;
         }
@@ -4389,7 +4399,7 @@ public class Parser {
                 }
             }
             Declaration prevd = null;
-            for (Declaration d : declList) {
+            for (Declaration d : globalDeclList) {
                 if (!target.equals(global) && d.type != null && d.type.javaName != null && d.type.javaName.length() > 0) {
                     // when "target" != "global", the former is a package where to output top-level classes into their own files
                     String shortName = d.type.javaName.substring(d.type.javaName.lastIndexOf('.') + 1);
--- a/src/main/java/org/bytedeco/javacpp/tools/Parser.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Parser.java
@@ -2331,8 +2331,10 @@ public class Parser {
                 break;
             }
         }
-        if (tokens.get().match("&&") || (context.javaName == null && localNamespace > 0) || (info != null && info.skip)) {
-            // This is a rvalue function, or a member function definition or specialization, skip over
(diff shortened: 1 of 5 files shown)


saudet:
Looks good, but let's make this optional so it doesn't break backward compatibility, probably with some new boolean flag like Info.friendify or something, that would work like Info.enumerate works for enums, but that activates your new code here instead.

HGuillemet:
It only adds new mappings. So the only way to break something is if the new mapping needs some `Info` to be mapped correctly. But I guess this will be the case for some presets, like it is for Pytorch with `GenericDict`.

Looking at how `Info` is working, I understand I have to write something [like this](https://github.com/bytedeco/javacpp/blob/232dcb0abf431330f70548c9920c8ae259d9d1f9/src/main/java/org/bytedeco/javacpp/tools/Parser.java#L3904-L3906).
Or shall we do something more generic and make`InfoMap.getFirst(null)` always act as defaults for more specific `InfoMap.get(cppName)` ?

saudet:
> It only adds new mappings. So the only way to break something is if the new mapping needs some `Info` to be mapped correctly. But I guess this will be the case for some presets, like it is for Pytorch with `GenericDict`.

Like you said yourself, the ADL rules are complicated, so if we enable this by default this is going to break something somewhere.

> Looking at how `Info` is working, I understand I have to write something [like this](https://github.com/bytedeco/javacpp/blob/232dcb0abf431330f70548c9920c8ae259d9d1f9/src/main/java/org/bytedeco/javacpp/tools/Parser.java#L3904-L3906). Or shall we do something more generic and make`InfoMap.getFirst(null)` always act as defaults for more specific `InfoMap.get(cppName)` ?

Yes, something like that.

HGuillemet:
If you define a global setting : `new Info().mapFriends()`, thus enabling the mapping for all friend functions.
And then an Info targetting a specific friend function but concerning some other setting, you have to also specify `.mapFriends()` or the function won't be mapped. That's not critical though.

saudet:
Right, let's not break consistency and/or backward compatibility for small things like that.

src/main/java/org/bytedeco/javacpp/tools/Parser.java
```diff
@@ -3660,7 +3669,16 @@ boolean group(Context context, DeclarationList declList) throws ParserException
         } else if (info != null && info.flatten) {
             info.javaText = decl.text;
         }
-        declList.add(decl);
+        if (declList.add(decl)) {
+            for (Declaration dd: declList2.nonMemberDeclarations) {
+                // Only add the friend functions that we know will be visible through ADL.
+                for (int i=0; i<dd.declarator.parameters.declarators.length; i++)
+                    if (dd.declarator.parameters.declarators[i].type.cppName.equals(type.cppName)) {
+                        globalDeclList.add(dd);
```
saudet, line 3677: Am I missing something or can this not be done without modifying DeclarationList? It seems like we should be able to check for type.friend here and there, resetting as necessary to pass it as a normal static function, and do what's needed without having to keep a list of "nonMemberDeclarations".

saudet: Also, why do we need to put this in the "global" list? We can put static methods anywhere in any class...

HGuillemet: The `nonMemberDeclarations` list is here to postpone the addition of the friend functions after the addition of the whole group AND if the addition of the group worked (`declList.add` returned `true`). Two reasons for this : 
1. if we directly add the friend functions when they are parsed, we end up with template types not replaced by java types. Apparently it's only when the whole group can be successfully added that we have the context info needed to replace the template types. Maybe there is a better way, I'm far from mastering how the parser handles templates.
2. if we add the friend function before the group to the global decl list, it will follow the class comment declaration, if any, and the comment will be attached to the friend function instead of correcty copied to the class file.

Concerning whether to translate the friend functions as static function of the global class or of the class they are friend of: both are possible, but I found more logical to make them methods of the global class. For example a function or operator combining 2  instances of 2 classes could be friend of both. Also many friend functions are also declared outside the class, to allow normal lookup instead of ADL only, in this case they end up in the global class.  There are already some equals, notEquals, or shiftLeft operators in the global class for this reason. Generally speaking friend functions are global functions, not member of the class. They appear in the class body only to be able to access private and protected members. Finally the user will often make a static import of the global class, allowing to not prefix the method like in `equals(iterator, end)`, instead of `named_module_iterator.equals(iterator, end)`.

saudet: Well, to make this more consistent, I think friend functions should appear in the class where they are declared. If they are also supposed to be part of the global namespace, they will appear as such as normal non-friend static functions, too, right?

saudet: Besides, it makes it easier to pair them with non-static methods like this, and also makes it clearer to users that these two methods are related:
```java
public class module_iterator extends Pointer {
...
@Namespace public static native @Name("operator !=") boolean notEquals(@Const @ByRef module_iterator a, @Const @ByRef module_iterator b);
@Override public boolean equals(Object o) { return o != null && o instanceof module_iterator && !notEquals(this, o); }
...
}
```
I'm not saying you need to do that right away, but if we start moving function declarations around, it makes it harder to bring them back later like this, and that's what needs to be done anyway because that's usually the intent of the original C++ code when declaring friends functions like that.

HGuillemet: From cppreference.com: 

> Names introduced by [friend](https://en.cppreference.com/w/cpp/language/friend) declarations within a non-local class X become members of the innermost enclosing namespace of X, but they do not become visible to ordinary [name lookup](https://en.cppreference.com/w/cpp/language/lookup) (neither [unqualified](https://en.cppreference.com/w/cpp/language/unqualified_lookup) nor [qualified](https://en.cppreference.com/w/cpp/language/qualified_lookup)) unless a matching declaration is provided at namespace scope, either before or after the class definition. Such name may be found through [ADL](https://en.cppreference.com/w/cpp/language/adl) which considers both namespaces and classes.

So a friend function belongs to the enclosing namespace, not to the class it is friend of. If the function was supposed to be a class member, it would have been declared static and not friend in C++.

In the current global `torch.java`, without this patch, there is already 14 `notEquals`, 5 `equals`, 25 `shiftLeft`, etc... and many other operators overload. So the most consistent seems to put the one defined as friend there too. Also I think your example about paired operators has little chance to happen: why `==` would have been defined as a member function and `!=` as a friend one ?

In fact the best user-friendly, java-ish, solution would be turn all operators overloads as instance methods, be them declared as such, as global not friend functions, or as friend functions. We would need to patch the generator. Is that possible ? One major problem is the case of global not-friend function, since the declaration can appear anywhere, far from the class definition. That would also break some existing code but that may be worth. And concerning non-operator friend functions, we would keep them as static methods (of global, which I think correspond the best to the C++ semantic, or of the friend class. The important is that the method is mapped).

saudet: > So a friend function belongs to the enclosing namespace, not to the class it is friend of. If the function was supposed to be a class member, it would have been declared static and not friend in C++.

Technically, yes, but like I said, if that function is meant to be used outside the class, it's going to have a static declaration there as well. The reason friends function exist in the first place is because there is no way to declare functions like `operator <<(ostream& os, ...)` as member functions. It's just a hack they came up with. But the intention is usually to treat them like member functions.

> In the current global `torch.java`, without this patch, there is already 14 `notEquals`, 5 `equals`, 25 `shiftLeft`, etc... and many other operators overload. So the most consistent seems to put the one defined as friend there too. Also I think your example about paired operators has little chance to happen: why `==` would have been defined as a member function and `!=` as a friend one ?

If the functions that are friended also have declarations in the global namespace, JavaCPP will continue to map there as well. You haven't made any changes in that respect, right? If the C++ API uses friend functions for some things but not others, that's a problem with original C++ API, not with JavaCPP. Do you have a concrete example that you're worried about?

Why would they declare `operator ==` as a member function and `operator !=` as a friend function indeed. Like I said, that's a problem with the original C++ API. Mapping both of them to `Object.equals()` seems fine though.

> In fact the best user-friendly, java-ish, solution would be turn all operators overloads as instance methods, be them declared as such, as global not friend functions, or as friend functions. We would need to patch the generator. Is that possible ? One major problem is the case of global not-friend function, since the declaration can appear anywhere, far from the class definition. That would also break some existing code but that may be worth. And concerning non-operator friend functions, we would keep them as static methods (of global, which I think correspond the best to the C++ semantic, or of the friend class. The important is that the method is mapped).

Sure, we can do a lot of things, but what I try to do with JavaCPP is not get into twisted situations that don't have a clear mapping from C++. Moving functions around like that just makes it harder for everyone to figure out what is going on. If we want something that's actually designed for Java or Scala, like @sbrunk is doing, we'd have to design manually the whole API from scratch anyway.

HGuillemet: > Technically, yes, but like I said, if that function is meant to be used outside the class, it's going to have a static declaration there as well.

Not if the function is an operator. Since operator are used with ADL.

HGuillemet: >  JavaCPP will continue to map there as well. You haven't made any changes in that respect, right? 

I didn't but if we follow your idea, won't we end with 2 mapping for the function? a static function when we encounter the friend declaration, and a global static one when we encounter the declaration outside the class.

HGuillemet: > Do you have a concrete example that you're worried about?

Let take two examples in Pytorch:

1. `module_iterator`, an instance of the template `torch::jit::slot_iterator_impl` defined in `jit/api/module.h`. `!=` is missing, but `++` is mapped as an instance method in `ModuleIterator.java`.
Because they are defined like this:
```c++
namespace torch {
namespace jit {

template <typename Policy>
struct slot_iterator_impl {
  slot_iterator_impl& operator++() {
    next_valid();
    return *this;
  }
  friend inline bool operator!=(
      const slot_iterator_impl<Policy>& a,
      const slot_iterator_impl<Policy>& b) {
    return (a.cursors_.empty() != b.cursors_.empty());
  } 
}
}
}
```
The developer chose to implement `++` as an instance method and `!=` as a friend method. `!=` is clearly meant to be visible. He could have implemented `!=` as an instance method. Would have been better for us but that wouldn't change anything for the C++ user that writes `i != end` in both cases. No need to declare it in the enclosing namespace too. If he did impement `!=` as an instance method, it would have been mapped as a Java instance method. If he added a declaration in enclosing namespace, it would have been mapped as static in `global.torch`. Never as a static method of `module_iterator`.

2. `TypeIdentifier`, defined in `c10/util/typeid.h`
```c++
class C10_API TypeIdentifier final
    : public at::IdWrapper<TypeIdentifier, c10::util::type_index> {
public:
  friend std::ostream& operator<<(std::ostream& stream, TypeIdentifier typeId);
  friend constexpr bool operator<(TypeIdentifier lhs, TypeIdentifier rhs);
};

// Allow usage in std::map / std::set
inline constexpr bool operator<(TypeIdentifier lhs, TypeIdentifier rhs) {
  return lhs.underlyingId() < rhs.underlyingId();
}
inline std::ostream& operator<<(
    std::ostream& stream,
    caffe2::TypeIdentifier typeId) {
  return stream << typeId.underlyingId();
}

```
`<<` is implemented as a friend function. Like you pointed it out, there was no choice here. 
`<` too, but the reason apparently here is that it needs to be redeclared at the namespace scope in order to be visible by `std::map` when a `TypeIdentifier` is used as a key. `<<` is also redeclared, for good measure I guess.
Both functions are thus mapped in `global.torch`. No need for this PR here.  No static functions in `TypeIdentifier`.

> Why would they declare operator == as a member function and operator != as a friend function indeed. Like I said, that's a problem with the original C++ API. Mapping both of them to Object.equals() seems fine though.

How would you map the friend function to `Object.equals() ` ?

> Sure, we can do a lot of things, but what I try to do with JavaCPP is not get into twisted situations that don't have a clear mapping from C++.

I agree, that's what I think sticking with the C++ semantic is best.

>  like @sbrunk is doing,

I missed that. What is he working on ?

sbrunk: > I missed that. What is he working on ?

I'm working on an idiomatic Scala API on top the the PyTorch bindings: https://storch.dev/

HGuillemet: Cool. Have you seen [this discussion about Pytorch 2](https://discuss.pytorch.org/t/pytorch-2-and-the-c-interface/168034) ?

sbrunk: Thanks, that's pretty interesting. The most interesting thing, the answer to your second question remains open though.

saudet: > Not if the function is an operator. Since operator are used with ADL.

Sure, of course we can. In C++ we can declare the same function any number of times. PyTorch deliberately doesn't define them in global namespace as well because they want us to think of them as sort-of-member function, I guess. So, why not take the hint? It's easier for JavaCPP that way too.

> I didn't but if we follow your idea, won't we end with 2 mapping for the function? a static function when we encounter the friend declaration, and a global static one when we encounter the declaration outside the class.

Sure, that's what it looks like in C++ too, so why is it a problem if we map them 1-to-1 with duplicates? It's the easiest thing to do and it works. If it's not broken, don't fix it!

> How would you map the friend function to Object.equals() ?

Something like I've done for multiple inheritance around here:
https://github.com/bytedeco/javacpp/blob/master/src/main/java/org/bytedeco/javacpp/tools/Parser.java#L3386
That all ends up in Declaration.text. We can *literally* put anything we want there.

HGuillemet: > > How would you map the friend function to Object.equals() ?
> 
> Something like I've done for multiple inheritance around here: https://github.com/bytedeco/javacpp/blob/master/src/main/java/org/bytedeco/javacpp/tools/Parser.java#L3386 That all ends up in Declaration.text. We can _literally_ put anything we want there.

Ok, so with this trick we could map as instance methods all operators declared as friends and taking an instance of the class or struct as its first argument. Can we agree on that ?

saudet: > Ok, so with this trick we could map as instance methods all operators declared as friends and taking an instance of the class or struct as its first argument. Can we agree on that ?

Sure, that sounds fine, although apart from equals(), I don't think we gain much by doing so...?

HGuillemet: Ok, I implemented this: map the friend function to a `private static native` method of the class the function is friend of, and add a Java wrapper as an instance method. It's what you had in mind, right ? 
I had to force the use of ADL instead of namespace lookup with a `@Namespace` annotation without argument or it didn't compile. So now all friend functions accepting at least 1 argument of the class it is friend of are accessible as instance method. 

Example:
```Java
  public boolean notEquals(StringView rhs) { return notEquals(this, rhs); }
  private static native @Namespace @Cast("bool") @Name("operator !=") boolean notEquals(@Const @ByRef StringView lhs, @Const @ByRef StringView rhs);

  public Pointer shiftLeft(Pointer os) { return shiftLeft(os, this); }
  private static native @Namespace @Cast("std::ostream*") @ByRef @Name("operator <<") Pointer shiftLeft(@Cast("std::ostream*") @ByRef Pointer os, @Const @ByRef StringView dt);
```

Do you agree with this solution ? Any remark before I repost a PR ?

HGuillemet: When a library developer implements an operator, (s)he usually has the choice between implementing it as a friend function or as an instance method. It doesn't change anything for the C++ user.

The good thing about this solution is that, with it, it doesn't change anything for the Java user either.

saudet: > Do you agree with this solution ? Any remark before I repost a PR ?

Yes, that looks exactly like what I had in mind. You can update this PR if you want, GitHub keeps all the commits either way. Thanks!

src/main/java/org/bytedeco/javacpp/tools/Parser.java
```diff
@@ -2512,8 +2517,34 @@ boolean function(Context context, DeclarationList declList) throws ParserExcepti
...
+                }
+                if (foundThis)
+                    decl.text += "public "+dcl.type.javaName+" "+dcl.javaName+"("+argList+") { "+(dcl.type.javaName.equals("void") ? "" : "return ")+dcl.javaName+"("+staticArgList+"); }\n";
+                else
+                    mapFriends = false;
+            }
+
             // skip over non-const function within const class
-            if (!decl.constMember && context.constName != null) {
+            if (type.friend && !mapFriends || !decl.constMember && context.constName != null) {
```
saudet, line 2547: To make sure we don't have any regressions, let's keep the skip condition for friend above instead.

saudet: I've pushed proposed changes in that direction. Can you give it a try? If it works, I think this is good to merge.

HGuillemet:
1. The wrapper instance method is added just as a string to the static method declaration. We thus bypass any chance to detect if the method already exists. This case happens in pytorch: in `ivalue.h`, an `equals` member function is defined and then a friend `operator ==`. Do you see a simple way to detect this ? If we add a true declaration for the instance method, will it work ? And can we give priority to the existing member function ?
2. The Info inheritance system is not well suited for this kind of setting: if we set a global `new Info(null).frendly()`, each time an info is created that apply to a function, the `friendly` setting is silently reset to false and the friend function is not parsed. What about changing the `Info(null)` so that it becomes a true set of default values for all new Info ? There is little change it would break anything. A system of cascaded default value like in CSS would be nice too but then we would break much more.

saudet:
> 1. The wrapper instance method is added just as a string to the static method declaration. We thus bypass any chance to detect if the method already exists. This case happens in pytorch: in `ivalue.h`, an `equals` member function is defined and then a friend `operator ==`. Do you see a simple way to detect this ? If we add a true declaration for the instance method, will it work ? And can we give priority to the existing member function ?

To take care of that, you just have to make sure their Declaration.signature is the same. Duplicates are removed here:
https://github.com/bytedeco/javacpp/blob/master/src/main/java/org/bytedeco/javacpp/tools/DeclarationList.java#L174

> 2. The Info inheritance system is not well suited for this kind of setting: if we set a global `new Info().frendly()`, each time an info is created that applies to a function, the `friendly` setting is silently reset to false and the friend function is not parsed. What about adding a true set of default values for all new Info ? For instance, we could add this to `InfoMap`:

We can already do that, just keep a "Info defaults" objects somewhere, and then create your `new Info(defaults)...` from it. You can also add any number of helper methods such as add() if you really want it. What's wrong with that?

> Nothing broken since the current style would be unchanged.

It's good that nothing breaks, but still, I don't see the point of maintaining this kind of code. It doesn't seem to add any value. Please prove me wrong.

HGuillemet:
> To take care of that, you just have to make sure their Declaration.signature is the same. Duplicates are removed here:
> https://github.com/bytedeco/javacpp/blob/master/src/main/java/org/bytedeco/javacpp/tools/DeclarationList.java#L174

Ok, so we need to create a separate `Declaration` for the instance method with the proper signature. The declaration that will be kept is the one appearing first in the C++ source I guess. It's a bit random but at least it will compile. The situation should not  often happen. Additional commit to come.


> It's good that nothing breaks, but still, I don't see the point of maintaining this kind of code. It doesn't seem to add any value. 

Turning on new features like `enumerate` and `friendly` is supposed to be common for all new presets. So you'll asks users to write:
```
Info def = new Info().enumerate().friendly();
infoMap.put(new Info(def).cppNames(...).pointerTypes(...))
       .put....
```
While you could allow them to write with less noise, like they are used to:
```
Info.default.enumerate().friendly();
infoMap.put(new Info(...).pointerTypes(...))
       .put....
```
By adding  just one more line to maintain in Info.java and changing another one:
```
static final public Info def = new Info();
public Info() { this(def); }
```
> Please prove me wrong.

Done ? :satisfied:

HGuillemet:
> By adding just one more line to maintain in Info.java and changing another one.

Not counting the lines in parser where we fetch `Info(null)` that could be removed !

But this change of the info system, if any, can wait. I have a related issue to point out and there may be a better alternative to this default info.

HGuillemet:
Committed. I also added a fix to bypass the `inaccessible` check in `group` for friends.
Tested successfully on Pytorch presets, although I had to manually `skip` some operators of `DictIterator` for some reason I failed to identified related to a template that couldn't be instantiated. If you can have a look, I'll be interested to understand with your help. I will push a PR of the modified Pytorch presets once this one is merged.

PR #650, opened 2023-01-25
Minor simplifications & typo

HGuillemet:
Minor simplifications, fixes, typos to the parser, mostly suggested by my IDE.
No functional changes.
Pick the ones you want to merge.

An optimization that could significantly improve the speed of parsing is to replace all these string concatenations by the use of `StringBuilder`. But we'll lose a bit in lisibility. So I'm not sure you'd want that.

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/tools/Declaration.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Declaration.java
@@ -32,4 +32,8 @@ class Declaration {
     boolean abstractMember = false, constMember = false, inaccessible = false,
             incomplete = false, function = false, variable = false, comment = false, custom = false;
     String signature = "", text = "";
+
+    public String toString() {
+        return text;
+    }
 }
--- a/src/main/java/org/bytedeco/javacpp/tools/Parser.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Parser.java
@@ -30,10 +30,12 @@ import java.nio.file.Files;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.Properties;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
@@ -129,12 +131,12 @@ public class Parser {
     }
 
     void containers(Context context, DeclarationList declList) throws ParserException {
-        List<String> basicContainers = new ArrayList<String>();
+        List<String> basicContainers = new ArrayList<>();
         for (Info info : infoMap.get("basic/containers")) {
             basicContainers.addAll(Arrays.asList(info.cppTypes));
         }
         for (String containerName : basicContainers) {
-            LinkedHashSet<Info> infoSet = new LinkedHashSet<Info>();
+            LinkedHashSet<Info> infoSet = new LinkedHashSet<>();
             infoSet.addAll(leafInfoMap.get("const " + containerName));
             infoSet.addAll(leafInfoMap.get(containerName));
             for (Info info : infoSet) {
@@ -236,7 +238,7 @@ public class Parser {
                     dcl.definition.text = "\n" + dcl.definition.text;
                     decl.declarator = dcl;
                 }
-                LinkedHashSet<Type> typeSet = new LinkedHashSet<Type>();
+                LinkedHashSet<Type> typeSet = new LinkedHashSet<>();
                 typeSet.addAll(Arrays.asList(firstType, secondType, indexType, valueType));
                 typeSet.addAll(Arrays.asList(containerType.arguments));
                 for (Type type : typeSet) {
@@ -616,7 +618,7 @@ public class Parser {
         if (!tokens.get().match('<')) {
             return null;
         }
-        List<Type> arguments = new ArrayList<Type>();
+        List<Type> arguments = new ArrayList<>();
         for (Token token = tokens.next(); !token.match(Token.EOF); token = tokens.next()) {
             Type type = type(context);
             arguments.add(type);
@@ -655,7 +657,7 @@ public class Parser {
                 break;
             }
         }
-        return arguments.toArray(new Type[arguments.size()]);
+        return arguments.toArray(new Type[0]);
     }
 
     Type type(Context context) throws ParserException {
@@ -664,7 +666,7 @@ public class Parser {
 
     Type type(Context context, boolean definition) throws ParserException {
         Type type = new Type();
-        List<Attribute> attributes = new ArrayList<Attribute>();
+        List<Attribute> attributes = new ArrayList<>();
         for (Token token = tokens.get(); !token.match(Token.EOF); token = tokens.get()) {
             if (token.match("::")) {
                 Info info = infoMap.getFirst(type.cppName, false);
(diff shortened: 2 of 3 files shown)


src/main/java/org/bytedeco/javacpp/tools/Parser.java
```diff
@@ -787,7 +789,7 @@ Type type(Context context, boolean definition) throws ParserException {
             } else if (token.match(Token.ENUM, Token.EXPLICIT, Token.EXTERN, Token.INLINE, Token.CLASS, Token.FINAL,
                                    Token.INTERFACE, Token.__INTERFACE, Token.MUTABLE, Token.NAMESPACE, Token.STRUCT, Token.UNION,
                                    Token.TYPENAME, Token.REGISTER, Token.THREAD_LOCAL, Token.VOLATILE)) {
-                token = tokens.next();
+                tokens.next();
```
saudet, line 792: That sounds like it could change the behavior. Let's keep those kinds of changes out.

saudet: Or maybe not...

src/main/java/org/bytedeco/javacpp/tools/Parser.java
```diff
@@ -4054,7 +4054,7 @@ void declarations(Context context, DeclarationList declList) throws ParserExcept
             declList.infoIterator = null;
             declList.spacing = null;
             do {
-                if (map != null && declList.infoIterator != null && declList.infoIterator.hasNext()) {
+                if (map != null && declList.infoIterator != null) {
```
saudet, line 4057: What about this one? Why does it remove the check? I guess we can kind of infer that it's always going to be true, but I'd rather leave it there...

src/main/java/org/bytedeco/javacpp/tools/Parser.java
HGuillemet: The returned token is not used indeed.

src/main/java/org/bytedeco/javacpp/tools/Parser.java
HGuillemet: `hasNext()` is always true here, see the while clause. And at first iteration `infoIterator` is null.

src/main/java/org/bytedeco/javacpp/tools/Declaration.java
```diff
@@ -32,4 +32,8 @@ class Declaration {
     boolean abstractMember = false, constMember = false, inaccessible = false,
             incomplete = false, function = false, variable = false, comment = false, custom = false;
     String signature = "", text = "";
+
+    public String toString() {
+        return text;
+    }
```
HGuillemet, line 38: This was to ease debugging.

src/main/java/org/bytedeco/javacpp/tools/Parser.java
```diff
@@ -655,7 +657,7 @@ Type[] templateArguments(Context context) throws ParserException {
                 break;
             }
         }
-        return arguments.toArray(new Type[arguments.size()]);
+        return arguments.toArray(new Type[0]);
```
HGuillemet, line 660: All these kind of changes follow this recommendation: "In older Java versions, using a pre-sized array was recommended, as the reflection call necessary to create an array of proper size was quite slow.
However, since late updates of OpenJDK 6, this call was intrinsified, making the performance of the empty array version the same, and sometimes even better, compared to the pre-sized version. Also, passing a pre-sized array is dangerous for a concurrent or synchronized collection as a data race is possible between the size and toArray calls. This may result in extra nulls at the end of the array if the collection was concurrently shrunk during the operation."

src/main/java/org/bytedeco/javacpp/tools/Parser.java
```diff
@@ -1021,7 +1023,7 @@ Type type(Context context, boolean definition) throws ParserException {
         if (context.cppName != null && type.javaName.length() > 0) {
             String cppName = type.cppName;
             String groupName = context.cppName;
-            int template2 = groupName != null ? groupName.lastIndexOf('<') : -1;
+            int template2 = groupName.lastIndexOf('<');
```
HGuillemet, line 1026: `groupName` is never null, see `if` clause 3 lines above.

src/main/java/org/bytedeco/javacpp/tools/Parser.java
```diff
@@ -1932,7 +1934,7 @@ Attribute attribute(boolean explicit) throws ParserException {
             brackets = false;
             tokens.next();
         }
-        if (!brackets && count <= 0) {
+        if (!brackets && count == 0) {
```
HGuillemet, line 1937: `count` is 0 or 1, see 5 line above.

src/main/java/org/bytedeco/javacpp/tools/Parser.java
```diff
@@ -3423,7 +3424,6 @@ boolean group(Context context, DeclarationList declList) throws ParserException
             } else if (context.namespace != null && context.javaName == null) {
                 decl.text += "@Namespace(\"" + context.namespace + "\") ";
             }
-            decl.type = new Type(name);
```
HGuillemet, line 3426: `decl.type` is re-assigned 2 lines below.

src/main/java/org/bytedeco/javacpp/tools/Parser.java
```diff
@@ -4129,7 +4129,7 @@ void declarations(Context context, DeclarationList declList) throws ParserExcept
         // for comments at the end without declarations
         String comment = commentBefore() + (tokens.get().match(Token.EOF) ? tokens.get().spacing : "");
         Declaration decl = new Declaration();
-        if (comment != null && comment.length() > 0) {
```
HGuillemet, line 4132: `comment` is never null, see 2 lines above

src/main/java/org/bytedeco/javacpp/tools/Parser.java
```diff
@@ -4208,22 +4208,17 @@ public File[] parse(File outputDirectory, String[] classPath, Class cls) throws
         ClassProperties allProperties = Loader.loadProperties(cls, properties, true);
         ClassProperties clsProperties = Loader.loadProperties(cls, properties, false);
 
-        List<String> excludes = new ArrayList<>();
-        excludes.addAll(clsProperties.get("platform.exclude"));
-        excludes.addAll(allProperties.get("platform.exclude"));
+        Set<String> excludes = new HashSet<>(allProperties.get("platform.exclude"));
```
HGuillemet, line 4211: `allProperties` already contains what is in `clsProperties`. And use an `HashSet` instead of an ArrayList since `excludes` is only used for `.contains(...)`

src/main/java/org/bytedeco/javacpp/tools/Parser.java
```diff
@@ -4272,9 +4267,11 @@ public File[] parse(File outputDirectory, String[] classPath, Class cls) throws
...
         boolean objectify = false;
         for (Info info : infoList) {
-            objectify |= info != null && info.objectify;
-            if (info.javaText != null && info.javaText.startsWith("import")) {
-                text += info.javaText + "\n";
+            if (info != null) {
+                objectify |= info.objectify;
+                if (info.javaText != null && info.javaText.startsWith("import")) {
+                    text += info.javaText + "\n";
+                }
```
HGuillemet, line 4274: If `info` is ever null you'd get an NPE line 4276.

PR #652, opened 2023-02-14
Add default Info

HGuillemet:
See https://github.com/bytedeco/javacpp/pull/649#issuecomment-1426844810

[This lines](https://github.com/bytedeco/javacpp/blob/5f81c237cf1a2a6f2100dc39ae58d3c7d6dd07e1/src/main/java/org/bytedeco/javacpp/tools/Parser.java#L2332-L2333) and [this lines](https://github.com/bytedeco/javacpp/blob/5f81c237cf1a2a6f2100dc39ae58d3c7d6dd07e1/src/main/java/org/bytedeco/javacpp/tools/Parser.java#L3948-L3949) are not  useful anymore with this PR. But you may want to keep them for compatibility.

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/tools/Info.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Info.java
@@ -44,7 +44,7 @@ public class Info {
         return def;
     }
     public Info() { this(def); }
-    public Info(String... cppNames) { this.cppNames = cppNames; }
+    public Info(String... cppNames) { this(def); this.cppNames = cppNames; }
     public Info(Info i) {
         if (i == null) return;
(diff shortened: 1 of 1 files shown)


HGuillemet:
Do not merge this for the moment. There is an obscure side effect [here](https://github.com/bytedeco/javacpp/blob/5f81c237cf1a2a6f2100dc39ae58d3c7d6dd07e1/src/main/java/org/bytedeco/javacpp/tools/Parser.java#L1436-L1442):
when `enumerate` is true per default on an Info where `pointerTypes` has been set, `valueType`  here is set to true and we iterate over the (single) `valueTypes` instead of the `pointerTypes`.

saudet:
Right, Info.enumerate() shouldn't be set for "normal" types.

PR #654, opened 2023-02-17
Fix output path when provided classes have same length

HGuillemet:
If `getOutputPath` is provided 2 classes with same path length, `longest` and `shortest` are assigned the same value, and the returned path is one of the 2 full path (depending on their order in the input array).
This PR fixes that (+ a typo in comment).

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/tools/Builder.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Builder.java
@@ -481,9 +481,8 @@ public class Builder {
                     String resourceURL2 = Loader.findResource(classes[i], resourceName2).toString();
                     String packageURI2 = resourceURL2.substring(0, resourceURL2.lastIndexOf('/') + 1);
 
-                    String longest, shortest;
-                    if (packageURI2.length() > packageURI.length()) { longest = packageURI2; shortest = packageURI; }
-                    else { longest = packageURI; shortest = packageURI2; }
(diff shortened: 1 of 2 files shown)


src/main/java/org/bytedeco/javacpp/tools/Builder.java
```diff
@@ -476,13 +476,14 @@ File getOutputPath(Class[] classes, String[] sourcePrefixes) throws IOException
...
+                    // Use the longest common package name among all classes as default output path
                     String resourceName2 = '/' + classes[i].getName().replace('.', '/')  + ".class";
                     String resourceURL2 = Loader.findResource(classes[i], resourceName2).toString();
                     String packageURI2 = resourceURL2.substring(0, resourceURL2.lastIndexOf('/') + 1);
 
-                    String longest = packageURI2.length() > packageURI.length() ? packageURI2 : packageURI;
-                    String shortest = packageURI2.length() < packageURI.length() ? packageURI2 : packageURI;
+                    String longest, shortest;
+                    if (packageURI2.length() > packageURI.length()) { longest = packageURI2; shortest = packageURI; }
+                    else { longest = packageURI; shortest = packageURI2; }
```
saudet, line 486: Let's just replace `>` with `>=` or `<` with `<=` instead?

HGuillemet: As you wish.

Issue #651, opened 2023-02-13
Virtual inheritance


HGuillemet:
In the JNI code, the `address` field of Java objects is interpreted as a pointer to a C++ object using a C-style cast.
This always works when the effective class of the object is the class we are casting to. This works also in most cases where the object class derived. But not always.

First case where this cast is illegal is when the base class `B` is polymorphic and the derived class `D` inherits virtually from `B`:

Example 1:
```c++
#include <cstdio>

class B {
  int b = 1;
  public:
    virtual void f() { printf("%d\n", b); }
};

class D: public virtual B {
  int d = 2;
};

```
If we write in Java: `new D().f()` we get a segmentation fault.
To prevent this, we would need to replace the C-style cast by a `static_cast<B*>(ptr)` where `ptr` is a `D*`, but we never have a chance to. First time we enter C++ code during this call is in the JNI of method `f` and at this point the code is specific to `B` and knows nothing about `D`.

Example 2, we add this function to Example 1:
```c++
void g(B b) {
  b.f();
}
```
When I call from Java `global.g(new D())`, on my machine I get `2`, which is incorrect. Others may get segmentation fault. Anyway the C-style cast `(B*)` made by the JNI of `g` on its argument is illegal.

Another case where C-style cast is illegal is multiple inheritance.

Example 3, we add:
```c++
class B2 {
  int b2 = 3;
};

class D2: public B, public B2 {
  int d2 = 4;
};
```
Here, no segmentation fault nor incorrect output when we call `new D2().f()` or `global.g(new D2())`. Java doesn't support multiple inheritance and JavaCPP only keeps the first inheritance relationship. I doubt this is written in any specification, but in practice C-style casts works to up-cast to the first class. JavaCPP also produces automatically an `asB2()` instance method for `D2` that does a `static_cast<B2*>`.

My suggestions:
1. Drop the Java inheritance in case of C++ virtual inheritance of a polymorphic class. The Java inheritance is just useless and only causes segmentation faults or, worse, hard to diagnose erroneous results. I don't think there is a way to have Java inheritance works here. Is there ?
2. Add an automatic `asB()` method  in this case, just like it's done for multiple inheritance, and also manually by `Info` in Pytorch presets for `Module` subclasses.
[...]

PR #655, opened 2023-02-18
Copy declarations between classes

HGuillemet:
This PR adds a mechanism to copy declarations from one class to another.
It is used automatically when a virtual inheritance towards a polymorphic class, that we cannot transpose to Java, is detected (see #651).
We can also copy on-demand with a new `Info.copyFrom`. This is useful for instance when a C++ API forwards calls to another class. Example with Pytorch:
```Java
.put(new Info("torch::nn::Linear").copyFrom("torch::nn::LinarImpl::LinearImpl", "torch::nn::LinearImpl::forward"))
.put(new Info("torch::nn::Linear::forward").annotations("@Name(\"operator()\")"))        
```
copies the constructors and the forward method from `LinearImpl` to `Linear`, whatever their arguments are.
We also map the new forward method to the `operator()` that does the C++ forwarding.

There is already the `flatten` info that replaces the inheritance of a superclass to all its subclasses by a copy of the declarations. It's not really the declarations that are copied, but the whole java text of the class, after a string
replacement of the superclass name by the subclass name. The string replacement has good chance to break things like annotations, and there is no check for duplicate declarations after the copy. This PR keeps `flatten()` for compatibility but I'm not sure it's a good idea, because of its limits and because the code of Parser would be simpler if we merged this feature with the new copy mechanism. Is `flatten()` used (it is not in the current bytedeco presets) ? In which case ?

Limit: the headers must be parsed in order from superclasses to subclasses for the detection of polymorphic classes to work. Also the classes we copy from must be parsed before the classes we copy to.

Possible breakage: only in the case of virtual inheritance of a polymorphic class. The new style flattening will be triggered automatically.

Still work in progress because more tests are needed.
Also I didn't bother yet with indentation of copied declaration or with copy of comments attached to declarations.

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/tools/Declaration.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Declaration.java
@@ -26,7 +26,7 @@ package org.bytedeco.javacpp.tools;
  *
  * @author Samuel Audet
  */
-class Declaration {
+class Declaration implements Cloneable {
     Type type = null;
     Declarator declarator = null;
     boolean abstractMember = false, constMember = false, inaccessible = false,
@@ -35,5 +35,15 @@ class Declaration {
 
     public String toString() {
         return text;
+    }
+
+    @Override
+    public Declaration clone() {
+        try {
+            // Shallow copy is enough for the current usage
+            return (Declaration) super.clone();
+        } catch (CloneNotSupportedException e) {
+            throw new AssertionError();
+        }
     }
 }
--- a/src/main/java/org/bytedeco/javacpp/tools/Declaration.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Declaration.java
@@ -26,24 +26,33 @@ package org.bytedeco.javacpp.tools;
  *
  * @author Samuel Audet
  */
-class Declaration implements Cloneable {
+class Declaration {
     Type type = null;
     Declarator declarator = null;
     boolean abstractMember = false, constMember = false, inaccessible = false,
             incomplete = false, function = false, variable = false, comment = false, custom = false;
     String signature = "", text = "";
 
-    public String toString() {
-        return text;
+    public Declaration() {
     }
 
-    @Override
-    public Declaration clone() {
-        try {
-            // Shallow copy is enough for the current usage
-            return (Declaration) super.clone();
-        } catch (CloneNotSupportedException e) {
-            throw new AssertionError();
-        }
+    Declaration(Declaration src) {
+        // Shallow copy is enough for the current usage
+        type = src.type;
+        declarator = src.declarator;
+        abstractMember = src.abstractMember;
+        constMember = src.constMember;
+        inaccessible = src.inaccessible;
+        incomplete = src.incomplete;
+        function = src.function;
+        variable = src.variable;
+        comment = src.comment;
+        custom = src.custom;
+        signature = src.signature;
+        text = src.text;
+    }
+
+    public String toString() {
+        return text;
     }
 }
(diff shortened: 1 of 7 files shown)


HGuillemet:
#654 was merged into this PR by mistake.

saudet:
You're starting to change a lot of things here. Could you please make sure that the output of the current presets do not change?

src/main/java/org/bytedeco/javacpp/tools/Declaration.java
```diff
@@ -36,4 +36,14 @@ class Declaration {
     public String toString() {
         return text;
     }
+
+    @Override
+    public Declaration clone() {
+        try {
+            // Shallow copy is enough for the current usage
+            return (Declaration) super.clone();
```
saudet, line 44: Please create a copy constructor instead.

src/main/java/org/bytedeco/javacpp/tools/Info.java
```diff
@@ -124,6 +129,14 @@ public Info(Info i) {
     String cppText = null;
     /** Outputs the given code, instead of the result parsed from the declaration of C++ identifiers. */
     String javaText = null;
+    /** List of cpp names (class, struct, member function) the declarations of which we want to copy in this class or struct. */
+    List<String> copyFrom = null;
+    /** If not-null, declarations for these cpp names are copied.
+     * Set to non-null by InfoMap.normalizeCopy or for polymorphic classes. */
+    CopiedDeclarations copiedDeclarations = null;
+    /** Set by Parser when a class has a virtual member function or inherit from a polymorphic class. */
+    boolean polymorphic = false;
```
saudet, line 138: Info is meant to contain configuration values from the user. If those are not meant to be set by the user, please find another place for them.

src/main/java/org/bytedeco/javacpp/tools/Declaration.java
HGuillemet: Done.

src/main/java/org/bytedeco/javacpp/tools/Info.java
HGuillemet: Done, it's better like this (although `Info` is already used to store things not set by user (like constructors text)).

HGuillemet:
> You're starting to change a lot of things here. Could you please make sure that the output of the current presets do not change?

I will, but maybe after you review the last changes, and comment about the points below... and if you don't prefer to reject the whole thing :smile: 

The copy of declarations is used here for virtual inheritance and on-demand copy, but I believe it can be reused for other situations:
* flattening. See comment in first message above. Shall we replace current flattening with copy of declarations ?
* inheriting constructors using C++11 `using`. Currently the text of all constructors is saved in `Info` and copied to subclass declaring a `using` after a string replacement of class names, similarly to flattening. Besides possible failure of the string replacement, the problem is that we cannot copy, using the new mechanism, constructors that have been copied as text like this. For instance in Pytorch we won't be able to copy constructors from `MaxPool3dImpl` to `MaxPool3d` because `MaxPool3dImpl` declares a `using MaxPoolImpl<3, MaxPool3dImpl>::MaxPoolImpl`. We could store all constructor declarations in a map, to have them available for copy when a `using` is parsed, but then it's a big duplicate feature with the current storage of all constructor texts in Infos.
* when a container type is added for smart pointers, for copying methods from the contained class to the container class ?

src/main/java/org/bytedeco/javacpp/tools/Info.java
saudet: Right, it does store some stuff by default, but that can be overridden by the user, or at least that's the intention and if it doesn't allow it it's probably a bug.

saudet:
> I will, but maybe after you review the last changes, and comment about the points below... and if you don't prefer to reject the whole thing smile

As long as it doesn't break anything, I think it's fine to add something like that, sure. But unless you're prepared to do a lot of testing to make sure it doesn't break someone's code somewhere, let's not break things intentionally like you're suggesting for flatten, etc.

HGuillemet:
The needs behind this feature are partly covered by PR #671, now merged.
Such feature may still be useful, but I think it's best to implement it when we rewrite the Parser.

So are you ok to close this PR ?

PR #657, opened 2023-03-05
Fix namespaces and template arguments parsing.

HGuillemet:
At various places, the parser uses simple character searchs like `indexOf('<')`  to find templates arguments, or namespace components. This PR replaces this with more robust searchs that should work in all cases like nested templates, `operator<`, namespace-qualified template arguments etc...

Also add a couple of minor improvements.

More details below.

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/tools/Context.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Context.java
@@ -90,6 +90,7 @@ class Context {
         }
         List<String> names = new ArrayList<String>();
         String ns = namespace != null ? namespace : "";
+        String prefix = infoMap.normalize(cppName, false, true);
         while (ns != null) {
             String name = ns.length() > 0 ? ns + "::" + cppName : cppName;
             if (parameters != null && name.endsWith(parameters)) {
@@ -118,7 +119,6 @@ class Context {
             names.add(name);
 
             for (String s : usingList) {
-                String prefix = infoMap.normalize(cppName, false, true);
                 int i = s.lastIndexOf("::") + 2;
                 String ns2 = ns.length() > 0 ? ns + "::" + s.substring(0, i) : s.substring(0, i);
                 String suffix = s.substring(i);
--- a/src/main/java/org/bytedeco/javacpp/tools/Parser.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Parser.java
@@ -3200,9 +3200,8 @@ public class Parser {
         String cppName = dcl.type.cppName;
         String baseType = context.baseType;
         if (baseType != null) {
-            boolean noTemplate = Templates.strip(cppName).length() == cppName.length();
             String baseTypeStripped = Templates.strip(baseType);
-            if (noTemplate && cppName.startsWith(baseTypeStripped)) {
+            if (Templates.notExists(cppName) && cppName.startsWith(baseTypeStripped)) {
                 cppName = baseType + cppName.substring(baseTypeStripped.length());
             }
         }
--- a/src/main/java/org/bytedeco/javacpp/tools/Parser.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Parser.java
@@ -130,6 +130,48 @@ public class Parser {
         return text;
     }
 
+    // Remove template arguments from s, taking care of nested templates, default arguments (xxx<>), operator <=>, ->, etc...
+    private static Pattern templatePattern = Pattern.compile("<[^<>=]*>");
+    private static String templateStrip(String s) {
+        Matcher m;
+        do {
+            m = templatePattern.matcher(s);
+            s = m.replaceFirst("");
+        } while (!m.hitEnd());
+        return s;
+    }
+
+    private boolean noTemplate(String s) {
+        return templateStrip(s).length() == s.length();
+    }
+
+    // Split s at ::, but taking care of qualified template arguments
+    private static List<String> namespaceSplit(String s) {
+        String sTemplatesMasked = s;
+        for (;;) {
+            Matcher m = templatePattern.matcher(sTemplatesMasked);
+            if (m.find()) {
+                char[] c = new char[m.end() - m.start()];
+                Arrays.fill(c, '.');
+                sTemplatesMasked = sTemplatesMasked.substring(0, m.start()) + new String(c) + sTemplatesMasked.substring(m.end());
+            } else
+                break;
+        }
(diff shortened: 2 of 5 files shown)


src/main/java/org/bytedeco/javacpp/tools/Context.java
```diff
@@ -90,6 +90,7 @@ String[] qualify(String cppName, String parameters) {
         }
         List<String> names = new ArrayList<String>();
         String ns = namespace != null ? namespace : "";
+        String prefix = infoMap.normalize(cppName, false, true);
```
HGuillemet, line 93: Moved this constant prefix out of the loops.

src/main/java/org/bytedeco/javacpp/tools/Parser.java
```diff
@@ -1700,8 +1727,7 @@ Declarator declarator(Context context, String defaultName, int infoNumber, boole
             if (context.namespace != null && localName.startsWith(context.namespace + "::")) {
                 localName = dcl.cppName.substring(context.namespace.length() + 2);
             }
-            int template = localName.lastIndexOf('<');
```
HGuillemet, line 1703: This failed to compute the right simpleName for `operator<<`

src/main/java/org/bytedeco/javacpp/tools/Parser.java
```diff
@@ -2296,15 +2322,8 @@ boolean function(Context context, DeclarationList declList) throws ParserExcepti
         if (info == null) {
             if (type.constructor) {
                 // get Info explicitly associated with all constructors
-                String name = dcl.cppName;
```
HGuillemet, line 2299: This failed for instance in Pytorch for `c10::ArrayRef<decltype(::c10::impl::ScalarTypeToCPPType<::c10::ScalarType::Bool>::t)>` by returning `ScalarTypeToCPPType` instead of `ArrayRef`

src/main/java/org/bytedeco/javacpp/tools/Parser.java
```diff
@@ -3665,15 +3680,8 @@ boolean group(Context context, DeclarationList declList) throws ParserException
                 inheritedConstructors += d.text;
             }
         }
-        String constructorName = originalName;
```
HGuillemet, line 3668: This failed for nested templates, for example in Pytorch with `c10::static_cast_with_inter_type<c10::complex<c10::Half>,c10::complex<double> >`

saudet:
Please test this on *all* presets to make sure it doesn't break anything.

HGuillemet:
I checked all presets, the PR introduces no changes in the result of parsing but for:
* pytorch (added missing `@NoDealloactor` due to nested template argument not correctly parsed)
* gym (nothing parsed ?)
* chilitags, scipy, tensorflow, tvm, qt, skia, systems, ngraph : could not compile them on my machine without significant efforts

Can we use the CI to check this last 8 presets ?

saudet:
Yes, but what we want to check is that the output of the Parser doesn't
change in unexpected ways. We can easily build everything in Docker as per
https://github.com/bytedeco/javacpp-presets/wiki/Build-Environments#linux-x86-and-x86_64

HGuillemet:
I could compile TensorFlow using Docker. It succeeds with master and fails with this PR, but I believe the parser does the right thing now and that the problem is with the preset config file:

When parsing `graph_def_builder.h`, the PR, for the `<tensorflow::gtl::ArraySlice<int>>` instance of `WithAttr` gives:
```
 public native @ByVal @Name("WithAttr<tensorflow::gtl::ArraySlice<int> >") Options WithAttr(@StringPiece BytePointer attr_name, @ArraySlice IntPointer value);
```
While master produces:
```
    public native @ByVal Options WithAttr(@StringPiece String attr_name, @ArraySlice IntPointer value);
```
The `@Name` annotation is missing, because [here](https://github.com/bytedeco/javacpp/blob/d23879af7a03a04c12b2374ae9d0850b9dda9d96/src/main/java/org/bytedeco/javacpp/tools/Parser.java#L1703-L1707) the parser fails to remove the nested template arguments.

However the JNI doesn't compile at this line:
```
        rptr = new tensorflow::GraphDefBuilder::Options(ptr->WithAttr<tensorflow::gtl::ArraySlice<int> >((tensorflow::StringPiece&)adapter0, (tensorflow::gtl::ArraySlice< int >&)adapter1));
```
with error:
```
error: cannot bind rvalue reference of type 'absl::Span<const int>&&' to lvalue of type 'tensorflow::gtl::ArraySlice<int>' {aka 'absl::Span<const int>'}
```
This seems to be related to the definition of the `ArraySlice` adapter.

HGuillemet:
This line seems to be missing from ngraph preset. Without it the parsed Java doesn't compile with the PR:
```
               .put(new Info("ngraph::element::from<ngraph::float16>").javaNames("fromNGraphFloat16"))
```

skia doesn't pass the cppbuild, even in docker, error:
```
bin/gn: /lib64/libc.so.6: version `GLIBC_2.18' not found (required by bin/gn)
```

Nothing is parsed in Scipy ?

The parsing result  doesn't differ for the remaining presets: chilitags, tvm, qt, systems.

saudet:
Ok, thanks for checking all this. We don't need to worry about TF 1.x and nGraph, those are obsolete and won't have any new releases, so I probably won't be making any for those either.

Yes, SciPy doesn't have a native API, that's normal.

I guess something changed with the build for Skia, but anyway we're not parsing its C++ API so no need to worry about it.

HGuillemet:
Good to merge ?

saudet:
Well, instead of having templateThis, templateThat, maybe we could have all that in a Template class just like you did with DocTag?

HGuillemet:
We can move static functions `templateStrip` and `noTemplate` only out of Parser to a `Templates` class, but this wouldn't gather all processing done in Parser about templates.
What about creating some `ParserUtils` class instead, to put those static functions and some more, like `namespaceSplit` and a couple of others to come in next PR ?

saudet:
There's no point in creating a class like ParserUtils... The point is, no other methods use that templatePattern you want to use. So, let's do something about that, like if you want to use more patterns and put them in static variables.

saudet:
Thanks! Looks good. Shall we merge?

HGuillemet:
Ok for me, you may want to add the missing line in ngraph and have a look at the error for tensorflow, so that all presets can be built. Or are those built already skipped ?

saudet:
They won't get built unless they get updated, and they won't get updated.

Issue #656, opened 2023-02-25
Virtualizing a class causes a method returning a String return a dangling pointer


HGuillemet:
The JNI code for method `name` of `Module` in Pytorch contains:
```
signed char* rptr;
StringAdapter< char > radapter(ptr->name());
rptr = radapter;
```
If we add an `Info` `virtualize()` on `Module`, the code becomes:
```
const signed char* rptr;
StringAdapter< char > radapter((dynamic_cast<JavaCPP_torch_0003a_0003ann_0003a_0003aModule*>(ptr) != NULL ? ((JavaCPP_torch_0003a_0003ann_0003a_0003aModule*)ptr)->name() : ptr->name()));
rptr = radapter;
```
The `const` modifier of `rptr`, when present or not, changes which overload of the implicit cast on the last line is applied. When absent, a copy of the string is performed. When present, no copy is performed and the returned pointer is dangling because the string it points to is freed at the end of the block.

Probably related to #374

saudet:
Does this happen only with PyTorch and only with Module? It's not possible to reproduce this with a smaller example?

HGuillemet:
Here is one:
```c++
#include <string>

class M {
  std::string name = "Bob";
  public:
    const std::string& n() const { return name; }
};
```
With `virtualize()`, parsed as:
```Java
    public native @Const @StdString @ByRef BytePointer n();
```
Generated JNI:
```c++
JNIEXPORT jobject JNICALL Java_virtopt_M_n(JNIEnv* env, jobject obj) {
    M* ptr = (M*)jlong_to_ptr(env->GetLongField(obj, JavaCPP_addressFID));
    if (ptr == NULL) {
        env->ThrowNew(JavaCPP_getClass(env, 9), "This pointer address is NULL.");
        return 0;
    }
    jlong position = env->GetLongField(obj, JavaCPP_positionFID);
    ptr += position;
    jobject rarg = NULL;
    const signed char* rptr;
    jthrowable exc = NULL;
    try {
        StringAdapter< char > radapter(ptr->n());
        rptr = radapter;
        jlong rcapacity = (jlong)radapter.size;
        void* rowner = radapter.owner;
        void (*deallocator)(void*) = rowner != NULL ? &StringAdapter< char >::deallocate : 0;
        if (rptr != NULL) {
            rarg = JavaCPP_createPointer(env, 10);
            if (rarg != NULL) {
[...]

PR #658, opened 2023-03-14
Narrow down context.virtualize at function level

HGuillemet:
See #656

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/tools/Parser.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Parser.java
@@ -2174,6 +2174,8 @@ public class Parser {
 
         int startIndex = tokens.index;
         Type type = type(context);
+        context = new Context(context);
+        context.virtualize &= type.virtual;
         Parameters params = parameters(context, 0, false);
         Declarator dcl = new Declarator();
         Declaration decl = new Declaration();
@@ -2563,7 +2565,7 @@ public class Parser {
             }
 
             // add @Const annotation only for const virtual functions
-            if (decl.constMember && type.virtual && context.virtualize) {
+            if (decl.constMember && context.virtualize) {
                 if (type.annotations.contains("@Const")) {
                     type.annotations = incorporateConstAnnotation(type.annotations, 2, true);
                 } else {
@@ -2572,7 +2574,7 @@ public class Parser {
             }
 
             // add @Virtual annotation on user request only, inherited through context
-            if (type.virtual && context.virtualize) {
+            if (context.virtualize) {
                 modifiers = "@Virtual" + (decl.abstractMember ? "(true) " : " ")
                           + (context.inaccessible ? "protected native " : "public native ");
             }
@@ -2627,7 +2629,7 @@ public class Parser {
                     first = false;
                     if (extraDecl != null) declList.add(extraDecl);
                 }
-                if (type.virtual && context.virtualize) {
+                if (context.virtualize) {
                     break;
                 }
             } else if (found && n / 2 > 0 && n % 2 == 0 && n / 2 > Math.max(dcl.infoNumber, dcl.parameters.infoNumber)) {
--- a/src/main/java/org/bytedeco/javacpp/tools/Parser.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Parser.java
@@ -2355,7 +2355,7 @@ public class Parser {
             }
         }
 
-        type = postDeclarator(context, decl, dcl, type);
+        type = functionAfter(context, decl, dcl, type);
         context = new Context(context);
         context.virtualize &= type.virtual;
 
@@ -2444,7 +2444,7 @@ public class Parser {
                 }
             }
 
-            type = postDeclarator(context, decl, dcl, type);
+            type = functionAfter(context, decl, dcl, type);
 
             if (tokens.get().match('{')) {
                 body();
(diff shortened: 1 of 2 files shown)


saudet:
Could you test this with all (non-obsolete) presets that use Info.virtualize() just to make sure nothing breaks?

HGuillemet:
When the `override` specifier is present, the member function is virtual, even if the `virtual` specifier is not present.
But the declarator is parsed before the `override` keyword, so we don't know the function is marked `override` yet, so we can fail to annotate the parameters correctly.
We could probably move the `override` parsing from [here](https://github.com/bytedeco/javacpp/blob/7a72aacb313e6b2bc488147a4cb2e89ef5eecac8/src/main/java/org/bytedeco/javacpp/tools/Parser.java#L2443) to [somewhere around here](https://github.com/bytedeco/javacpp/blob/7a72aacb313e6b2bc488147a4cb2e89ef5eecac8/src/main/java/org/bytedeco/javacpp/tools/Parser.java#L2229).

But the `override` keyword is optional and a function overriding a virtual function is virtual, even if not marked `virtual` nor `override`. Right ? So if we want to solve the issue rigorously, we would need to detect the actual overriding. This logic doesn't exist in Parser (but it does in Generator).

What do you think about this ?
This is not strictly related to this PR. What made me think about it is that the PR now ignores the override keyword when checking if `@Virtual` must be added, compared to master. I'll fix this, but how to do it depends on your answer about the issue above.

saudet:
Well that identifier only appears after the parameters, so the easiest way to fix this might be to call function() again recursively?

HGuillemet:
I didn't see an easy way to call `function` recursively, and since the declaration is already parsed a first time before the `for (int n = -2; n < Integer.MAX_VALUE; n++)` loop but stops after the parameters, I just pushed it a little further to include the post-parameter specifiers.

I re-checked all presets with this new commit.

No change in any parsing result but:

__caffe__: a parameter of non-virtual functions of virtualized class have lost their `@Const` annotation:
* constructors of *Solver (DoubleAdaDeltaSolver, DoubleAdamSolver...)
* loss, set_loss, param_propagate_down, set_param_propagate_down in DoubleLayer and FloatLayer. 
JNI compiles.

I guess the new behavior is correct ?

__qt__: same for QByteArray.fromStdString, QString.fromStdString and QString.fromStdWString

not tested: arpack-ng (could not compile with CentOS docker), skia (could not download)

saudet:
What are you trying to do exactly here? I don't see how that is supposed to pick up `override` after the parameters.

src/main/java/org/bytedeco/javacpp/tools/Parser.java
```diff
@@ -2440,32 +2444,8 @@ boolean function(Context context, DeclarationList declList) throws ParserExcepti
...
-                } else if (attr == null) {
-                    break;
-                }
-            }
-            if (tokens.get().match("->")) {
-                // auto type
-                tokens.next();
-                type = type(context);
-            }
+            type = postDeclarator(context, decl, dcl, type);
```
saudet, line 2447: And if the first call is enough, why do we need a second call here?

HGuillemet:
> What are you trying to do exactly here? I don't see how that is supposed to pick up `override` after the parameters.

`postDeclarator` returns a modified `Type` with `virtual` set if `override` is present. What makes you say it doesn't ?

The aim is to set `context.virtual` when `override` is present, not only when `virtual` is present, so that:
* when `declarator` is called in the `for (int n...` loop the proper annotations are set on parameters for virtualization
* the `@Virtual` annotation is added.

In master, `context.virtual` is true for all functions of a virtualized class, so the parameter annotations is always added (which caused the bug with `Module.name()` in Pytorch when Module is virtualized), and `@Virtual` is added correctly.

src/main/java/org/bytedeco/javacpp/tools/Parser.java
HGuillemet: `decl`, `dcl` and `type` are reset in the `for (int n...` loop. The whole declaration is reparsed at each iteration. 
 
And we need to swallow this part of the declaration before continuing the parsing.

src/main/java/org/bytedeco/javacpp/tools/Parser.java
```diff
@@ -2355,6 +2355,10 @@ boolean function(Context context, DeclarationList declList) throws ParserExcepti
             }
         }
 
+        type = postDeclarator(context, decl, dcl, dcl.type);
```
HGuillemet, line 2358: Maybe passing `type` instead of `dcl.type` is better ?

src/main/java/org/bytedeco/javacpp/tools/Parser.java
saudet: Hum, what happens if you move that in declarator(), just before `return dcl`, and only call it only when dcl.parameters is not null?

HGuillemet: `declarator` is called a lot of times, not only for functions, and this post-declarator part is function-specific.

HGuillemet: Could `declarator` be called from somewhere else than `function` and this post-declarator part is expected ?

saudet: Probably not, but there are other function-only related things in there because it gets called in a loop in function(). The const part, for example, gets picked up there too.

saudet: But we need that information before calling parameters() so I think that's not going to help anyways.

HGuillemet: Right. 
And the post-declarator set `constMember` which is in `decl`, not `dcl`.
Wha is the problem with putting this part of parsing in postDeclarator ?

src/main/java/org/bytedeco/javacpp/tools/Parser.java
saudet: Probably, does it change anything?

HGuillemet: No, they are the same at this line

PR #664, opened 2023-03-23
Add Info.defaultInsertable

HGuillemet:
Add a new Info to hint that a class cannot be default-constructed in containers.

WIthout this PR, some containers cannot be mapped to Java because the Parser will generate a `resize` method that is not supported.

Seen in pytorch for things like `std::vector<torch::OrderedDict<std::string,std::shared_ptr<torch::nn::Module>>::Item>` or `std::vector<torch::nn::AnyValue>`.

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/tools/Info.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Info.java
@@ -65,6 +65,7 @@ public class Info {
         base = i.base;
         cppText = i.cppText;
         javaText = i.javaText;
+        defaultInsertable = i.defaultInsertable;
     }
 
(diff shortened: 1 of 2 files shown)


saudet:
I'm pretty sure that's already covered by prefixing the cppNames with "const" like those here: https://github.com/bytedeco/javacpp-presets/blob/master/pytorch/src/main/java/org/bytedeco/pytorch/presets/torch.java#L2118

HGuillemet:
Ah, I missed that.
Adding `const` does more than just skipping `resize`: it also skips `put` and `erase`, and replaces `get` by `at`, but it should do the trick for the current need in Pytorch.
You can put this PR on hold.

PR #668, opened 2023-03-31
Add share Info

HGuillemet:
`shared_ptr<C>` are handled by specifying this info:
```Java
new Info("std::shared_ptr<C>")
  .annotations("@SharedPtr")
  .valueTypes("@Cast({\"\", \"std::shared_ptr<C>\"}) C")
  .pointerTypes("C")
```
which results in the ability to use Java class `C` indifferently to represent a native `C` or a `shared_ptr<C>`. 
This works well, but this is incomplete, because if a `c = new C()` is instantiated, and then passed to a native function taking a `shared_ptr<C>`, and then the instance is deallocated, either explicitly with `c.deallocate()`, or by a PointerScope, or by the GC, the native object will be freed, causing a SIGSEGV if the native library tries to access `c`. Thus ruining the point of using shared pointers. Currently we can add this Info on C constructor :
```Java
new Info("C::C").annotations("@NoDeallocator")
```
to avoid the SIGSEGV, but then instances of C will never be freed. We need to be able to call `make_shared` from Java-side.

This PR adds a new Info:
```Java
new Info("C").share()
```
that:
1. does the same things than
```Java
new Info("std::shared_ptr<C>")
  .annotations("@SharedPtr")
  .valueTypes("@Cast({\"\", \"std::shared_ptr<C>\"}) C")
  .pointerTypes("C")
```
2. changes the constructor from the usual:
```Java
public C() { super((Pointer) null); allocate();
private native void allocate();
```
to:
```Java
public C() { super(makeShared()); }
@Namespace @SharedPtr @Name("SHARED_PTR_NAMESPACE::make_shared<C>") static private native C makeShared();
```

What do you think of this solution ?

An alternative, a bit more efficient, to makeShared, is to add a `@SharedPtr` annotation to `allocate`, and have the Generator output an alternate version of `allocate`, but that would need to patch Generator and its magic is a bit to dark for me.


saudet:
> An alternative, a bit more efficient, to makeShared, is to add a `@SharedPtr` annotation to `allocate`, and have the Generator output an alternate version of `allocate`, but that would need to patch Generator and its magic is a bit to dark for me.

Yes, we should be doing that. `allocate()` is just a normal JNI function, you can look as where `@NoDeallocator` gets used. That's probably where something like this should be. We might as well generalize this and check for `@Adapter`  instead and wrap make_shared in a static function of SharedPtrAdapter.

And with the Parser, we could let users provide an Info.annotations(...) on the constructor I guess.

HGuillemet:
I can try something in this direction.

Also I'm wondering about the relevance of generating the `@SharedPtr`, `@Cast` and `@PointerTypes` automatically in the parser from a `Info.share()`. In pytorch you wrote some variants: like [here](https://github.com/bytedeco/javacpp-presets/blob/b228e731377b66af8eee9a94d7af916408b52b68/pytorch/src/main/java/org/bytedeco/pytorch/presets/torch.java#L2399-L2401) and or [here](https://github.com/bytedeco/javacpp-presets/blob/b228e731377b66af8eee9a94d7af916408b52b68/pytorch/src/main/java/org/bytedeco/pytorch/presets/torch.java#L2472-L2474).
What was the logic there and is that implementable in the Parse ?

HGuillemet:
For handling `const` in `SharedPtrAdapter`, I think that we could instantiate the template only for non-const types, and add a constructor accepting a const with something like this:
```c++
template<class T> class SharedPtrAdapter {
public:
    typedef SHARED_PTR_NAMESPACE::shared_ptr<T> S;
    typedef SHARED_PTR_NAMESPACE::shared_ptr<const T> SC;
    SharedPtrAdapter(const T* ptr, size_t size, void* owner) : ptr((T*)ptr), size(size), owner(owner),
            sharedPtr2(owner != NULL && owner != ptr ? *(S*)owner : S((T*)ptr)), sharedPtr(sharedPtr2) { }
    SharedPtrAdapter(const S& sharedPtr) : ptr(0), size(0), owner(0), sharedPtr2(sharedPtr), sharedPtr(sharedPtr2) { }
    SharedPtrAdapter(      S& sharedPtr) : ptr(0), size(0), owner(0), sharedPtr(sharedPtr) { }
    SharedPtrAdapter(const SC& sharedPtr) : ptr(0), size(0), owner(0), sharedPtr2(std::const_pointer_cast<T>(sharedPtr)), sharedPtr(sharedPtr2) { }
    SharedPtrAdapter(const S* sharedPtr) : ptr(0), size(0), owner(0), sharedPtr(*(S*)sharedPtr) { }
```
For functions taking either `std::shared_ptr<T>` or `std::shared_ptr<const T>` we can already pass the `std::shared_ptr<T>` resulting from a cast from the adapter.
For functions returning either `std::shared_ptr<T>` or `std::shared_ptr<const T>`, we could now create an adapter in both cases.

Am I wrong ?

saudet:
Should like something that could work...

HGuillemet:
Could you review that ? 
I will generalize to other smart pointers if it looks ok.

We must also find a way to have explicit cast `asX()` keep the existing `shared_ptr`.

src/main/java/org/bytedeco/javacpp/tools/Generator.java
```diff
@@ -1350,33 +1350,37 @@ boolean classes(boolean handleExceptions, boolean defineAdapters, boolean conver
...
             out.println("template<class T> class SharedPtrAdapter {");
             out.println("public:");
-            out.println("    typedef SHARED_PTR_NAMESPACE::shared_ptr<T> S;");
-            out.println("    SharedPtrAdapter(const T* ptr, size_t size, void* owner) : ptr((T*)ptr), size(size), owner(owner),");
-            out.println("            sharedPtr2(owner != NULL && owner != ptr ? *(S*)owner : S((T*)ptr)), sharedPtr(sharedPtr2) { }");
-            out.println("    SharedPtrAdapter(const S& sharedPtr) : ptr(0), size(0), owner(0), sharedPtr2(sharedPtr), sharedPtr(sharedPtr2) { }");
-            out.println("    SharedPtrAdapter(      S& sharedPtr) : ptr(0), size(0), owner(0), sharedPtr(sharedPtr) { }");
-            out.println("    SharedPtrAdapter(const S* sharedPtr) : ptr(0), size(0), owner(0), sharedPtr(*(S*)sharedPtr) { }");
-            out.println("    void assign(T* ptr, size_t size, void* owner) {");
+            out.println("    typedef typename SHARED_PTR_NAMESPACE::remove_const<T>::type TU;");
```
HGuillemet, line 1353: Either these `remove_const`, `add_const`, or we must ensure in `adapterInformation` that `valueTypeName` has no const (for all adapters ??).

saudet:
> Could you review that ?
> I will generalize to other smart pointers if it looks ok.

Well, did you test it with all presets that use `@SharedPtr`? The problem I see with this is that it's probably going to break a lot of things...

src/main/java/org/bytedeco/javacpp/tools/Parser.java
```diff
@@ -1006,7 +1023,8 @@ Type type(Context context, boolean definition) throws ParserException {
         }
         if (info != null && info.annotations != null) {
             for (String s : info.annotations) {
-                type.annotations += s + " ";
+                if (!type.annotations.contains(s))
```
HGuillemet, line 1026: Are there annotations which could legitimately be duplicated ?

src/main/java/org/bytedeco/javacpp/tools/Parser.java
```diff
@@ -3166,7 +3185,7 @@ boolean typedef(Context context, DeclarationList declList) throws ParserExceptio
                         info.valueTypes(info.pointerTypes != null ? info.pointerTypes : new String[] {typeName});
                         info.pointerTypes("PointerPointer");
                     } else if (info.pointerTypes == null) {
-                        info.pointerTypes(typeName);
```
HGuillemet, line 3169: A fix not directly related to this PR.
`typeName` is a cpp name. We need a java name for `pointerTypes`. Or did I miss something ?

HGuillemet:
> > Could you review that ?
> > I will generalize to other smart pointers if it looks ok.
> 
> Well, did you test it with all presets that use `@SharedPtr`? The problem I see with this is that it's probably going to break a lot of things...

Not yet. I did for Pytorch. I'd like to know if you see something wrong first.
We cannot test easily for sure like we can for Parser-only changes by comparing parser outputs.
What may break existing presets using @SharedPtr but not the new share info is the new definition of the adapter. But theoretically it could be able to handle const variations transparently.

saudet:
Right, so what about moving the development of that to the PyTorch presets? We can add adapters there like this:
https://github.com/bytedeco/javacpp-presets/blob/master/opencv/src/main/resources/org/bytedeco/opencv/include/opencv_adapters.h
And then when everything works for PyTorch, you can start moving things back to SharedPtrAdapter.

saudet:
We don't need to make any modifications to the core of JavaCPP to do most of what you're proposing for that too. There's a lot we can move to the presets for PyTorch. Why do you want to modify everything here?

HGuillemet:
Don't you think the ability to create shared objects from Java is a good thing that can benefit to other presets ?  Or are you talking about something else ?

saudet:
Like I said, we can already provide functions like that without modifying anything from the core of JavaCPP. It would be nice if the Generator could pick up annotations on allocate() and call a predetermined static function from the adapter, but you're not doing this here, and the rest of what you're doing can be done without modifying the core of JavaCPP.

HGuillemet:
> It would be nice if the Generator could pick up annotations on allocate() and call a predetermined static function from the adapter,

Did you have a look at the changes ? It does exactly that : when `@Adapter` is detected on allocate(), it instantiate an adapter instead of the object directly. No need to add a static function to the adapter. Unless I didn't understand what you meant.

HGuillemet:
Here is an example of what an allocate with @SharedPtr looks like:
```c++
JNIEXPORT void JNICALL Java_shared_M_allocate__(JNIEnv* env, jobject obj) {
    jthrowable exc = NULL;
    try {
        SharedPtrAdapter< ns::M > radapter(new ns::M(), 1, NULL);
        ns::M* rptr = radapter;
        jlong rcapacity = (jlong)radapter.size;
        void* rowner = radapter.owner;
        void (*deallocator)(void*) = rowner != NULL ? &SharedPtrAdapter< ns::M >::deallocate : 0;
        JavaCPP_initPointer(env, obj, rptr, rcapacity, rowner, deallocator);
    } catch (...) {
        exc = JavaCPP_handleException(env, 8);
    }

    if (exc != NULL) {
        env->Throw(exc);
    }
}
```

saudet:
I see, we don't need to change SharedPtrAdapter to wrap on allocate(). So let's start by merging in that modification only.

HGuillemet:
How do you add an annotation on `allocate` from an Info ? We need something more in the Parser to allow that.
What seemed the most logical for me is an info on the whole class, meaning "the objects of this class must be managed by smart pointers, don't create unmanaged instances". And if we add an info to annotate all `allocate` methods, this info might as well automatize the addition of the `@Cast` `@PointerType` `@SharedPtr` formula on the `shared_ptr` type. You seem to agree with that [above](https://github.com/bytedeco/javacpp/pull/668#issuecomment-1493186921), but maybe I misunderstood.

The PR doesn't do much more than that. I'll move out the modification of the adapter about const for now. There is also a modification of `asX` to replace the `static_cast` by a `static_pointer_cast` and return a managed object. Because we don't want to create unmanaged instances. This will be still more useful in the PR about explicit cast since it'll use these `asX` explicit cast to jump over virtual inheritances.

Now, as said [above](https://github.com/bytedeco/javacpp/pull/668#issuecomment-1494170875), I did this as a proof of concept for shared pointer but it needs to be generalized. 
We can either:
* generalize to smart pointers. Replacing `Info.share` with something like `info.managed("shared_ptr")`, or create other infos like `Info.weak`, `Info.unique`
* generalize to any adapter. In this case we may need to add functions to adapters to implement what is done by `asX`.

SInce I haven't spent time on understanding the roles of other adapters yet and hwo they work, I have no opinion about this.

saudet:
We probably don't need to add a new field to Info for any of that. Using the Info.annotations given for constructors should be sufficient. As for Type.shared and shared_ptr fields, from what I can see, that's only useful to add the annotation hacks for PyTorch that are very specific to how it makes use of std::shared_ptr, so like I keep telling you, that belongs in the presets of PyTorch, unless we figure out some way to generalize that, yes, sure, but instead of trying to do everything in this single PR, let's first merge things that clearly make sense.

src/main/java/org/bytedeco/javacpp/tools/Parser.java
HGuillemet: I let this change in. 
I think it fixes a bug but I can revert the change, or move it to a 1-line PR if you want.

HGuillemet:
> that's only useful to add the annotation hacks for PyTorch that are very specific to how it makes use of std::shared_ptr

Sorry but my experience with presets are limited. What exactly is special in how Pytorch uses std::shared_ptr ?

I'm just realizing that it can just work without any info, I thought that the pointerTypes and annotations(@Cast...@SharedPtr) was mandatory for the merge of native class X and shared_ptr<X> into a single java class X could work.

src/main/java/org/bytedeco/javacpp/tools/Generator.java
```diff
@@ -2401,16 +2401,28 @@ void parametersBefore(MethodInformation methodInfo) {
 
     String returnBefore(MethodInformation methodInfo) {
         String returnPrefix = "";
+        String[] typeName =
+            methodInfo.allocator || methodInfo.arrayAllocator ? cppTypeName(methodInfo.cls) :
+                (methodInfo.returnRaw ? new String[]{""}
+                    : cppCastTypeName(methodInfo.returnType, methodInfo.annotations));
```
saudet, line 2407: Is this doing anything different from the existing code in master?

src/main/java/org/bytedeco/javacpp/tools/Parser.java
saudet: Yes, your fix should be fine, thanks.

src/main/java/org/bytedeco/javacpp/tools/Generator.java
HGuillemet: You mean the move of typeName definition at the beginning of the method ? I guess I did that for consistency with returnAfter. But I can rechange for something closer to master.

saudet: Well, if it's not supposed to change anything, let's not "fix it" if it's _ain't_ broken. The original code looks fine to me.

HGuillemet: Done

src/main/java/org/bytedeco/javacpp/tools/Generator.java
```diff
@@ -2838,22 +2853,32 @@ void returnAfter(MethodInformation methodInfo) {
...
+                    } else if (methodInfo.arrayAllocator) {
+                        out.println("&JavaCPP_" + mangle(methodInfo.cls.getName()) + "_deallocateArray);");
+                        arrayDeallocators.index(methodInfo.cls);
+                    } else {
+                        out.println("&JavaCPP_" + mangle(methodInfo.cls.getName()) + "_deallocate);");
+                        deallocators.index(methodInfo.cls);
+                    }
                 }
                 if (virtualFunctions.containsKey(methodInfo.cls)) {
-                    typeName = cppTypeName(methodInfo.cls);
```
saudet, line 2856: Is there a reason to remove that line here?

saudet: I guess we could also move that line up a bit, and revert the lines at the beginning of this method too?

HGuillemet: This assignment is now done line 2804

HGuillemet: We need the type name for initializing the adapterInfo. That's why typeName is set at the beginning of the method for constructors too.

saudet: Right, I realize that, but if we can move that line just after `if (methodInfo.allocator || methodInfo.arrayAllocator)` we can revert the changes for its initial assignment, right?

HGuillemet: `adapterInfo` is initialized above, line 2809. We need the correct valueTypeName at this point. The initial assignment for `typeName` doesn't consider the `methodInfo.allocator` case. Or do I miss something ?

Something that you can double check is the case `methodInfo.arrayAllocator`. I didn't test this code path.

src/main/java/org/bytedeco/javacpp/tools/Generator.java
```diff
@@ -2838,22 +2853,32 @@ void returnAfter(MethodInformation methodInfo) {
...
                         methodInfo.cls.isAnnotationPresent(NoDeallocator.class) ||
                         methodInfo.method.isAnnotationPresent(NoDeallocator.class);
-                out.print(indent + "JavaCPP_initPointer(env, obj, rptr, rcapacity, rptr, ");
-                if (noDeallocator) {
-                    out.println("NULL);");
-                } else if (methodInfo.arrayAllocator) {
-                    out.println("&JavaCPP_" + mangle(methodInfo.cls.getName()) + "_deallocateArray);");
-                    arrayDeallocators.index(methodInfo.cls);
+                if (adapterInfo != null) {
+                    out.println(indent + typeName[0] + " rptr = radapter;");
```
saudet, line 2860: If you use typeName[1] as well like `typeName[0] + " rptr" + typeName[1] + " = radapter;"`, does it not work?

HGuillemet: Yes you must be right, although it seems that `typeName[1]` could differ from `""` only if the object we are allocating is a subclass of `FunctionPointer`

PR #671, opened 2023-04-09
Add explicitUpcast

HGuillemet:
This PR addresses issue #651.
A first attempt to solve it was PR #655.
But I find the solution brought by this PR better because:
* it introduces much less changes in the Parser
* the new code mostly runs when the new Info `explicitUpcast` is activated, so there is little chance of breaking anything
* we keep the Java inheritance relationship

How it works:
* new Info `explicitUpcast` is used to mark those classes that need a `static_cast` for converting a pointer of a subclass to a pointer of this class.
* Cast `asB()` methods are added for all marked classes `B` and all their direct subclasses.
* wrapper are created for all instance native methods of marked class and all native methods taking an instance of marked classes as an argument. For instance:
```Java
class B {
  public void f() { asB()._f(); }
  private native @Name("f") void _f();
  
  public B asB() { return this }
}

class D extends B {
  @Override public B asB() {  return asB(this); }
  public static native @Name("static_cast<B*>") B asB(B b);
}

class X {
  static public void g(B b) { _g(b.asB()); }
  static private native @Name("g") void _g(B b);
}
```
* Since the native method doesn't have its original name (`_f` instead of `f`), in order to still support virtualization, we add an optional parameter `javaName` to annotation `@Virtual` that tells the generator the name of the java callback method (`f` in this example). This optional parameter is added for all virtualized wrapped method.
* we trace polymorphic classes and virtual inheritance, in order to raise a warning (could be an error) if virtual inheritance to a polymorphic class is detected and if the parent polymorphic class doesn't have `explicitUpcast`. This feature could be removed if you prefer to minimize the code changes.

`asX` explicit casts were already used to map multiple inheritance. This PR proposes a version of `asX` that can takes care of smart pointers (could be in a separate PR), by using adapters and replacing the `static_cast` by a `static_pointer_cast`. However this still need improvements because:
1. it only supports `shared_ptr`
2. Without a specific info to mark classes managed by smart pointers (as initially proposed in PR #668) we have no easy way to detect when to generate the adapter-version of this method. Current non-satisfactory way is to test for the presence of an annotation on the constructor of one of the class. Do you have a better idea ? 

Finally, to improve readability and avoid code repetitions I introduced method `removeAnnotations`. This yields many line changes but no functional changes.

I'll compare parsing result of existing presets to detect breakages if this PR is to be merged.

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/annotation/Virtual.java
+++ b/src/main/java/org/bytedeco/javacpp/annotation/Virtual.java
@@ -21,4 +21,5 @@ public @interface Virtual {
     /** Pure (abstract) or not. */
     boolean value() default false;
     boolean subclasses() default true;
+    String javaName() default "";
 }
--- a/src/main/java/org/bytedeco/javacpp/annotation/Virtual.java
+++ b/src/main/java/org/bytedeco/javacpp/annotation/Virtual.java
@@ -21,5 +21,5 @@ public @interface Virtual {
     /** Pure (abstract) or not. */
     boolean value() default false;
     boolean subclasses() default true;
-    String javaName() default "";
+    String method() default "";
 }
--- a/src/main/java/org/bytedeco/javacpp/tools/Context.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Context.java
@@ -47,13 +47,13 @@ class Context {
         inaccessible = c.inaccessible;
         beanify = c.beanify;
         objectify = c.objectify;
+        upcast = c.upcast;
         virtualize = c.virtualize;
         variable = c.variable;
         infoMap = c.infoMap;
         templateMap = c.templateMap;
         usingList = c.usingList;
         namespaceMap = c.namespaceMap;
-        upcast = c.upcast;
     }
 
     String namespace = null;
@@ -66,13 +66,13 @@ class Context {
     boolean inaccessible = false;
     boolean beanify = false;
     boolean objectify = false;
+    boolean upcast = false;
     boolean virtualize = false;
     Declarator variable = null;
     InfoMap infoMap = null;
     TemplateMap templateMap = null;
     List<String> usingList = null;
     Map<String,String> namespaceMap = null;
-    boolean upcast = false;
 
     /** Return all likely combinations of namespaces and template arguments for this C++ type */
     String[] qualify(String cppName) {
--- a/src/main/java/org/bytedeco/javacpp/tools/Context.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Context.java
@@ -53,7 +53,7 @@ class Context {
         templateMap = c.templateMap;
         usingList = c.usingList;
         namespaceMap = c.namespaceMap;
-        explicitUpcast = c.explicitUpcast;
+        upcast = c.upcast;
     }
 
     String namespace = null;
@@ -72,7 +72,7 @@ class Context {
     TemplateMap templateMap = null;
     List<String> usingList = null;
     Map<String,String> namespaceMap = null;
-    boolean explicitUpcast = false;
+    boolean upcast = false;
 
     /** Return all likely combinations of namespaces and template arguments for this C++ type */
     String[] qualify(String cppName) {
--- a/src/main/java/org/bytedeco/javacpp/tools/Context.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Context.java
@@ -53,6 +53,7 @@ class Context {
         templateMap = c.templateMap;
         usingList = c.usingList;
         namespaceMap = c.namespaceMap;
+        explicitUpcast = c.explicitUpcast;
     }
 
     String namespace = null;
@@ -71,6 +72,7 @@ class Context {
     TemplateMap templateMap = null;
     List<String> usingList = null;
     Map<String,String> namespaceMap = null;
+    boolean explicitUpcast = false;
 
     /** Return all likely combinations of namespaces and template arguments for this C++ type */
     String[] qualify(String cppName) {
--- a/src/main/java/org/bytedeco/javacpp/tools/Parser.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Parser.java
@@ -134,7 +134,7 @@ public class Parser {
 
     private static String explicitCastMethodName(String javaName) {
         String shortName = javaName.substring(javaName.lastIndexOf('.') + 1);
-        return "as" + shortName;
+        return "as" + Character.toUpperCase(shortName.charAt(0)) + shortName.substring(1);
     }
 
     private static String joinList(List<?> list, String separator) {
--- a/src/main/java/org/bytedeco/javacpp/tools/Parser.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Parser.java
@@ -3313,8 +3313,8 @@ public class Parser {
            some cast function on the adapter instead of static_pointer_cast. */
         String annotation = getConstructorAdapter(to);
         if (annotation != null)
-            res += "    @Namespace public static native " + annotation + " @Name(\"SHARED_PTR_NAMESPACE::static_pointer_cast<" + to.cppName + ">\") "
-                + to.javaName + " " + ecmn + "(@Cast({\"\", \"SHARED_PTR_NAMESPACE::shared_ptr<" + from.cppName + ">\"}) " + annotation + " " + from.javaName + " pointer);\n";
+            res += "    @Namespace public static native " + annotation + " @Name(\"SHARED_PTR_NAMESPACE::static_pointer_cast<" + to.cppName + ", " + from.cppName + ">\") "
+                + to.javaName + " " + ecmn + "(" + annotation + " " + from.javaName + " pointer);\n";
         else
             res += "    @Namespace public static native @Name(\"static_cast<" + to.cppName + "*>\") "
                 + to.javaName + " " + ecmn + "(" + from.javaName + " pointer);\n";
--- a/src/main/java/org/bytedeco/javacpp/tools/Parser.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Parser.java
@@ -2559,7 +2559,7 @@ public class Parser {
             if (context.virtualize && !type.annotations.contains("@Virtual")) {
                 ArrayList<String> attrs = new ArrayList<>();
                 if (decl.abstractMember) attrs.add("value = true");
-                if (needWrapper) attrs.add("javaName = \"" + dcl.javaName+'"');
+                if (needWrapper) attrs.add("method = \"" + dcl.javaName+'"');
                 modifiers = "@Virtual" + (attrs.size() == 0 ? " " : ('(' + joinList(attrs, ", ") + ") "))
                           + (context.inaccessible ? "protected native " : "public native ");
             }
--- a/src/main/java/org/bytedeco/javacpp/tools/Parser.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Parser.java
@@ -95,7 +95,6 @@ public class Parser {
     InfoMap leafInfoMap = null;
     TokenIndexer tokens = null;
     String lineSeparator = null;
-    HashSet<String> polymorphicClasses = new HashSet<>(); // Contains Java names
 
     String translate(String text) {
         Info info = infoMap.getFirst(text);
@@ -3367,18 +3366,15 @@ public class Parser {
         boolean anonymous = !typedef && type.cppName.length() == 0, derivedClass = false, skipBase = false;
         if (type.cppName.length() > 0 && tokens.get().match(':')) {
             derivedClass = true;
-            boolean virtualInheritance = false;
-            boolean accessible = !ctx.inaccessible;
             for (Token token = tokens.next(); !token.match(Token.EOF); token = tokens.next()) {
+                boolean accessible = !ctx.inaccessible;
                 if (token.match(Token.VIRTUAL)) {
-                    virtualInheritance = true;
                     continue;
                 } else if (token.match(Token.PRIVATE, Token.PROTECTED, Token.PUBLIC)) {
                     accessible = token.match(Token.PUBLIC);
-                    continue;
+                    tokens.next();
                 }
                 Type t = type(context);
-                t.virtual = virtualInheritance;
                 Info info = infoMap.getFirst(t.cppName);
                 if (info != null && info.skip) {
                     skipBase = true;
@@ -3389,8 +3385,6 @@ public class Parser {
                 if (tokens.get().expect(',', '{').match('{')) {
                     break;
                 }
-                virtualInheritance = false;
-                accessible = !ctx.inaccessible;
             }
         }
         if (typedef && type.indirections > 0) {
@@ -3488,29 +3482,15 @@ public class Parser {
             }
             infoMap.put(info = new Info(type.cppName).pointerTypes(type.javaName));
         }
-
-        // Choose the base Java type: first base C++ type which is not flattened
-        // Keep the others in baseClasses to generate asX() casts.
-        // Detect virtual inheritance.
-        // If no C++ base type suits, use Pointer.
-        boolean polymorphic = false;
-        Type base = null;
+        Type base = new Type("Pointer");
         Iterator<Type> it = baseClasses.iterator();
         while (it.hasNext()) {
             Type next = it.next();
-            boolean nextPolymorphic = polymorphicClasses.contains(next.javaName);
-            polymorphic |= nextPolymorphic;
             Info nextInfo = infoMap.getFirst(next.cppName);
-            if (nextInfo != null && nextInfo.flatten)
-                continue;
-            if (nextPolymorphic && next.virtual && !next.explicitUpcast) {
-                // We can detect this only if the superclass is parsed before.
-                logger.warn(type.cppName + " virtually inherits from polymorphic class " + next.cppName +
-                    ". Consider adding an explicitUpcast Info on " + next.cppName + ".");
-            }
-            if (base == null) {
+            if (nextInfo == null || !nextInfo.flatten) {
                 base = next;
                 it.remove();
+                break;
             }
         }
         String casts = "";
@@ -3526,8 +3506,7 @@ public class Parser {
             casts += "    public " + type.javaName + ' ' + explicitCastMethodName(type.javaName) + "() { return this; }\n";
         }
 
-        if (base == null) base = new Type("Pointer");
-        else if (base.explicitUpcast) {
+        if (base.explicitUpcast) {
             casts += explicitCast(type, base, true);
         }
 
@@ -3629,7 +3608,6 @@ public class Parser {
                 pointerConstructor = false, abstractClass = info != null && info.purify && !ctx.virtualize,
                 allPureConst = true, haveVariables = false;
         for (Declaration d : declList2) {
-            polymorphic |= d.declarator != null && d.declarator.type != null && d.declarator.type.virtual;
             if (d.declarator != null && d.declarator.type != null && d.declarator.type.using && decl.text != null) {
                 // inheriting constructors
                 defaultConstructor |= d.text.contains("private native void allocate();");
@@ -3756,8 +3734,6 @@ public class Parser {
                 decl.custom = true;
             }
         }
-        if (polymorphic) polymorphicClasses.add(type.javaName);
-
         for (Declaration d : declList2) {
             if ((!d.inaccessible || d.declarator != null && d.declarator.type.friend)
                     && (d.declarator == null || d.declarator.type == null
--- a/src/main/java/org/bytedeco/javacpp/tools/Parser.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Parser.java
@@ -3311,21 +3311,23 @@ public class Parser {
            assuming that this annotation is @SharedPtr or some other adapter storing a share_ptr in owner.
            This is the only use case for now. We can generalize to any adapter if the need arises and call
            some cast function on the adapter instead of static_pointer_cast. */
-        if (constructorIsAnnotated(to))
-            res += "    @Namespace public static native @SharedPtr @Name(\"SHARED_PTR_NAMESPACE::static_pointer_cast<" + to.cppName + ">\") "
-                + to.javaName + " " + ecmn + "(@Cast({\"\", \"SHARED_PTR_NAMESPACE::shared_ptr<" + from.cppName + ">\"}) @SharedPtr " + from.javaName + " pointer);\n";
+        String annotation = getConstructorAdapter(to);
+        if (annotation != null)
+            res += "    @Namespace public static native " + annotation + " @Name(\"SHARED_PTR_NAMESPACE::static_pointer_cast<" + to.cppName + ">\") "
+                + to.javaName + " " + ecmn + "(@Cast({\"\", \"SHARED_PTR_NAMESPACE::shared_ptr<" + from.cppName + ">\"}) " + annotation + " " + from.javaName + " pointer);\n";
         else
             res += "    @Namespace public static native @Name(\"static_cast<" + to.cppName + "*>\") "
                 + to.javaName + " " + ecmn + "(" + from.javaName + " pointer);\n";
         return res;
     }
 
-    private boolean constructorIsAnnotated(Type t) {
+    private String getConstructorAdapter(Type t) {
         String cppName = t.cppName;
         List<String> split = Templates.splitNamespace(cppName);
         String constructorName = cppName + "::" + Templates.strip(split.get(split.size() - 1));
         Info info = infoMap.getFirst(constructorName);
-        return (info != null && info.annotations != null);
+        return info == null || info.annotations == null ? null :
+            joinList(Arrays.asList(info.annotations), " ");
     }
 
     boolean group(Context context, DeclarationList declList) throws ParserException {
--- a/src/main/java/org/bytedeco/javacpp/tools/Parser.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Parser.java
@@ -83,6 +83,7 @@ public class Parser {
         this.properties = p.properties;
         this.encoding = p.encoding;
         this.infoMap = p.infoMap;
+        this.explicitUpcasts = p.explicitUpcasts;
         Token t = p.tokens != null ? p.tokens.get() : Token.EOF;
         this.tokens = new TokenIndexer(infoMap, new Tokenizer(text, t.file, t.lineNumber).tokenize(), false);
         this.lineSeparator = p.lineSeparator;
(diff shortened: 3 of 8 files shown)


src/main/java/org/bytedeco/javacpp/tools/Parser.java
```diff
@@ -1031,6 +1056,9 @@ else if (cppNameSplit.size() > 1 && groupNameSplit.size() == 1)
             }
             type.javaName = context.shorten(type.javaName);
         }
+        type.explicitUpcast =
+            info != null && info.explicitUpcast
+                 || Templates.isSmartPointer(type.cppName) && type.arguments != null && type.arguments[0].explicitUpcast;
```
HGuillemet, line 1061: If you don't like this special case about smart pointers we can remove it. The presets will just have to add an `explicitCast` info on `std::shared_ptr<X>` in addition to the info on `X`.

saudet:
This looks like something reasonable, although instead of trying to work on unrelated changes all at once, let's at least separate all that into a couple of independent commits.

> * Since the native method doesn't have its original name (`_f` instead of `f`), in order to still support virtualization, we add an optional parameter `javaName` to annotation `@Virtual` that tells the generator the name of the java callback method (`f` in this example). This optional parameter is added for all virtualized wrapped method.

There must be a cleaner way of doing this...

> `asX` explicit casts were already used to map multiple inheritance. This PR proposes a version of `asX` that can takes care of smart pointers (could be in a separate PR), by using adapters and replacing the `static_cast` by a `static_pointer_cast`. However this still need improvements because:
> 
> 1. it only supports `shared_ptr`
> 
> 2. Without a specific info to mark classes managed by smart pointers (as initially proposed in PR [Add share Info #668](https://github.com/bytedeco/javacpp/pull/668)) we have no easy way to detect when to generate the adapter-version of this method. Current non-satisfactory way is to test for the presence of an annotation on the constructor of one of the class. Do you have a better idea ?

I'm not sure what you're asking here. Functions like static_pointer_cast return a new shared_ptr, but we don't want to do that here anyway. Simply casting the pointer is sufficient. If you want to create a new shared_ptr though, we can still use SharedPtrAdapter for that, just like you did in pull #668.

> I'll compare parsing result of existing presets to detect breakages if this PR is to be merged.

Please do!

HGuillemet:
> > * Since the native method doesn't have its original name (`_f` instead of `f`), in order to still support virtualization, we add an optional parameter `javaName` to annotation `@Virtual` that tells the generator the name of the java callback method (`f` in this example). This optional parameter is added for all virtualized wrapped method.
> 
> There must be a cleaner way of doing this...

Any idea ? Unless we accept of not making the thing transparent and ask the user to use the wrapped name of the method (`_f` or other) when overriding in Java, I don't see another way. This new option adds just one line of code in Generator.
> 
> > `asX` explicit casts were already used to map multiple inheritance. This PR proposes a version of `asX` that can takes care of smart pointers (could be in a separate PR), by using adapters and replacing the `static_cast` by a `static_pointer_cast`. However this still need improvements because:
> > 
> > 1. it only supports `shared_ptr`
> > 2. Without a specific info to mark classes managed by smart pointers (as initially proposed in PR [Add share Info #668](https://github.com/bytedeco/javacpp/pull/668)) we have no easy way to detect when to generate the adapter-version of this method. Current non-satisfactory way is to test for the presence of an annotation on the constructor of one of the class. Do you have a better idea ?
> 
> I'm not sure what you're asking here. Functions like static_pointer_cast return a new shared_ptr, but we don't want to do that here anyway. Simply casting the pointer is sufficient. If you want to create a new shared_ptr though, we can still use SharedPtrAdapter for that, just like you did in pull #668.

`static_pointer_cast` returns a new shared pointer to the same pointer, increasing the use count. That's what we need.
If we simply cast the pointer, we will create a java object containing a shared pointer or, if we use the adapter, an object containing a new shared pointer but with a reinitialised use count. In both case we will end up with a segmentation fault.
For unique and weak pointers I guess normal cast is fine, but I'm unsure.

I can move this part to another PR if you wish, but then the explicitUpcast feature would be incompatible with the use of  shared pointers.

saudet:
> If we simply cast the pointer, we will create a java object without a shared pointer or, if we use the adapter, an object containing a new shared pointer but with a reinitialised use count. In both case we will end up with a segmentation fault.

When not creating a new shared_ptr, we don't set the ownerAddress, so it won't get deallocated or anything, that's fine, it works.

saudet:
Right, anyway please use SharedPtrAdapter to create one, not static_pointer_cast.

HGuillemet:
We need both, don't we ? That's how it works currently in the PR.

saudet:
I don't see why we need static_pointer_cast. We can do exactly the same thing with SharedPtrAdapter.

saudet:
Or actually we'd need to extend SharedPtrAdapter and add a constructor with 2 types like the "possible implementation" here I guess: https://en.cppreference.com/w/cpp/memory/shared_ptr/pointer_cast#Version_1

saudet:
Anyway, like I keep telling you, stop trying to do everything at the same time. You don't need to have all of it working exactly like it needs to be before we can merge. Let's get one thing at a time working.

HGuillemet:
> Anyway, like I keep telling you, stop trying to do everything at the same time. You don't need to have all of it working exactly like it needs to be before we can merge. Let's get one thing at a time working.

The implement of asX and the explicitCast feature seems related to me, but no problem for splitting that in two PR. Anything else you want me to separate ?

saudet:
We should be able to use the adapter annotation from the constructor as is, that's what it's for. If you want to go into the details of all that and do a full POC before finalizing anything, that's fine, go ahead. I don't see where the problem is though. If you need to discuss this further, you'll need to provide more details about where you feel there is a problem.

HGuillemet:
> I don't see where the problem is though. If you need to discuss this further, you'll need to provide more details about where you feel there is a problem.

This PR currently generates either:
```
@Namespace public static native @Name("static_cast<B*>") B asB(D pointer);
```
or
```
@Namespace public static native @SharedPtr @Name("std::static_pointer_cast<B>") B asB(@Cast({"","std::shared_ptr<D>"}) @SharedPtr D pointer);
```
But, as you said, the second version could be generalized to something like:
```
@Namespace public static native @Adapter("AAA") @Name("AAA<B>::cast<D>") B asB(@Adapter("AAA") D pointer);
```
With the addition of a `cast` function in all adapters.
Did I understood well what you meant ?
It seems a good generalization to me. Would the cast do something meaningful for adapters others than sharedptr ?

Currently, the PR, to determine if the second version must be generated or the first one, looks for an annotation Info on the constructor of B or D. It looks clunky for me since:
* It should as well be B AND D. 
* B and/or D could be abstract and the annotated constructor may be on subclasses (that's the case for Pytorch, where `ReLUImpl` inherits from `Cloneable<ReLUImpl>` which virtually inherits from `Module`. The annotated constructors are those of `ReLUImpl` and `Module`.)
* We look for an info on the argument-less cppname of the constructor, but this might be on the fullname info. We could even find constructors with the annotations, and some without.
* This is not like a annotation introspection, the test is based on String.contains. And we should handle @SharedPtr-style shortcuts somehow.

But if we cannot find something better and not too complicated, we can live with that I guess.

saudet:
> With the addition of a cast function in all adapters.
> Did I understood well what you meant ?

Sort of, like I said, we should be able to add a constructor instead of a method. 

> This is not like a annotation introspection, the test is based on String.contains. And we should handle @SharedPtr-style shortcuts somehow.

We shouldn't need to introspect anything at all. Look, just try to make it work, if you're not able to make it work, we'll look at more concrete problems that you encounter when you encounter them. And like I keep telling you, why do you want to merge everything all at the same time? It makes it very hard to get anything done!

HGuillemet:
I asked you above if you want me to split off something else of this PR than the asX implementation.

saudet:
It's probably better that way yes, but if you want to have a complete POC that does absolutely everything you want, first, then by all means start working on that! As a first step, since I'm pretty sure we can do what you want to do with a new constructor to SharedPtrAdapter, please try to do that and see how that goes. Talking about every possible and impossible case isn't going to help.

saudet:
Sorry, I think I'm starting to understand what you're saying. Basically, we need something like a function call for asXXX()...  unless we start doing something special in Generator. But let's say we don't do that, how should the user be able to pass the name of that function call... Let me think about that

HGuillemet:
For me this PR is the POC. It works with Pytorch and I'm currently using it. It's the last one of the series of PR (but probably not split enough) for consolidating Pytorch presets. 

I realize I'm not very clear in my explanations. That could be interesting to generalize and pass an arbitrary function call for the cast. But:
* this information (how to cast an adapter of X to an adapter of Y)  seems specific to adapters, and not presets-specific, so the solutions of an added constructor or static function to the adapters seems better.
* is it worth generalizing ? can you imagine a use case for other adapters than sharedptr ?

Concerning the split of this PR, I can cut it into
- `explicitUpcast` info and the wrapping stuff
- detection of virtual inheritance for issuing the warning/error
- asXX implementation
- removeAnnotations utility function
- support for @Virtual 

Is that ok ?

saudet:
I like to consider one thing at a time, but if you're comfortable rewriting things, we can keep going in this thread, fixing one thing at a time, that's fine.

Let's see, about the static_pointer_cast function template, we can think of it as something that isn't specific to shared_ptr, it just currently doesn't apply to anything else than shared_ptr. So, what if we just use it whenever there is a user specified annotation on the constructor *and* that upcast mode is enabled? It wouldn't break anything, and we don't need to add anything about "shared_ptr" in the Parser and/or new knobs that are most likely never going to get used, for now anyway.

HGuillemet:
> I like to consider one thing at a time, but if you're comfortable rewriting things, we can keep going in this thread, fixing one thing at a time, that's fine.

Ok, then I'll just split off the detection of virtual inheritance for now. That's something optional that you can decide to merge or not, and that concerns separate lines of codes. The other features are more related to each others.

> 
> Let's see, about the static_pointer_cast function template, we can think of it as something that isn't specific to shared_ptr, it just currently doesn't apply to anything else than shared_ptr. So, what if we just use it whenever there is a user specified annotation on the constructor _and_ that upcast mode is enabled? It wouldn't break anything, and we don't need to add anything about "shared_ptr" in the Parser and/or new knobs that are most likely never going to get used, for now anyway.

I agree that we can postpone generalization until the need arises, if ever.

Well, `static_pointer_cast` IS specific to shared pointers, whatever you think of it as, and we need the `@SharedPtr` annotation in this line of code. So you'll have something about `shared_ptr` in Parser anyway. Is this such a problem ? `@SharedPtr` annotation is part of core JavaCPP. 
So since you'll have this bit of code specific to shared pointers, and you test for the presence of an annotation on the constructor, why not test if this annotation IS `@SharedPtr` ?

Let's sum up the implications of this `static_pointer_cast`:
1. it's needed when we want class *B* to be managed by shared pointer. The closest we have now to test this is the presence of a `@SharedPtr` annotation on B constructor. 
2. It may be helpful when the instance of D contains a shared pointer, in case the resulting B is passed to a function taking a `shared_ptr<B>`. We cannot really test this situation. Can we ?
3. It's armful if the instance of D has an owner which is `!= address` and is not a shared pointer. Since it'll be treated as a shared pointer by the adapter.

Can we assume that testing situation 1 is enough to exclude situation 3 ? I think so (although I'd prefer something to test on the class itself rather than on a constructor: an info or a `@SharedPtr` annotation).

We can check for `explicitUpcast` too to limit the risk, but this will exclude the following case: 
* D multiple-inherit from B1 and B2.
* B2 is managed by shared pointers.

That could happen. If we limit to `explicitUpcast`, this case may trigger segmentation faults due to pointers owned by multiple shared pointers.

HGuillemet:
I checked all presets. No change in parsing (besides `@Virtual(true)` now `@Virtual(value = true)`).

saudet:
> Well, `static_pointer_cast` IS specific to shared pointers, whatever you think of it as, and we need the `@SharedPtr` annotation in this line of code. So you'll have something about `shared_ptr` in Parser anyway. Is this such a problem ? `@SharedPtr` annotation is part of core JavaCPP. So since you'll have this bit of code specific to shared pointers, and you test for the presence of an annotation on the constructor, why not test if this annotation IS `@SharedPtr` ?

Many libraries use their own version of std::shared_ptr, such as OpenCV:
https://github.com/bytedeco/javacpp-presets/blob/master/opencv/src/main/resources/org/bytedeco/opencv/include/opencv_adapters.h
 
> Let's sum up the implications of this `static_pointer_cast`:
> 
>  1. it's needed when we want class _B_ to be managed by shared pointer. The closest we have now to test this is the presence of a `@SharedPtr` annotation on B constructor.

No, it's only needed when we want to *cast* explicitly some shared_ptr. That only happens in Parser for multiple inheritance or that upcast mode, and yes we only want to do that when `@SharedPtr` is on the constructor.

>  2. It may be helpful when the instance of D contains a shared pointer, in case the resulting B is passed to a function taking a `shared_ptr<B>`. We cannot really test this situation. Can we ?

How is that not covered by your upcast mode?

>  3. It's armful if the instance of D has an owner which is `!= address` and is not a shared pointer. Since it'll be treated as a shared pointer by the adapter.

Right, that's going to give a compiler error, so that's not a big issue.

> Can we assume that testing situation 1 is enough to exclude situation 3 ? I think so (although I'd prefer something to test on the class itself rather than on a constructor: an info or a `@SharedPtr` annotation).

Ideally, sure, but whatever we come up with is unlikely to be used by anyone at all, so let's not fix things that aren't broken.

> We can check for `explicitUpcast` too to limit the risk, but this will exclude the following case:
> 
> * D multiple-inherit from B1 and B2.

Sure, but that's alright for backward compatibility.
 
> * B2 is managed by shared pointers.

static_pointer_cast doesn't get used unless there is multiple inheritance, so like I keep saying, you'll need to explain a bit more what you're worried about here.

> That could happen. If we limit to `explicitUpcast`, this case may trigger segmentation faults due to pointers owned by multiple shared pointers.

Why? You mean we won't be changing the default behavior so that's a bad thing? There has been exactly zero issue reported about that, so there's no reason to fix what isn't broken.

HGuillemet:
> Many libraries use their own version of std::shared_ptr, such as OpenCV: https://github.com/bytedeco/javacpp-presets/blob/master/opencv/src/main/resources/org/bytedeco/opencv/include/opencv_adapters.h

Argh. I didn't realize that. I see `cv::Ptr` has its own `staticCast`, but that does a simple `std::static_pointer_cast`, which works bacause `cv::Ptr` inherits from `std::static_pointer_cast`.

So calling a constructor or static factory function on the adapter seems the best. We don't need to implement it on all adapters and let the compiler throws an error when it's needed and doesn't exist.
But we can also for now only support std::shared_ptr if you prefer.
 
> 
> > Let's sum up the implications of this `static_pointer_cast`:
> > 
> > 1. it's needed when we want class _B_ to be managed by shared pointer. The closest we have now to test this is the presence of a `@SharedPtr` annotation on B constructor.
> 
> No, it's only needed when we want to _cast_ explicitly some shared_ptr. That only happens in Parser for multiple inheritance or that upcast mode, and yes we only want to do that when `@SharedPtr` is on the constructor.

You need it when you need all instances of the class to contain a shared_ptr. And you put `@SharedPtr` on the constructor for this reason. Why do you argue on this point ?
 
> > 2. It may be helpful when the instance of D contains a shared pointer, in case the resulting B is passed to a function taking a `shared_ptr<B>`. We cannot really test this situation. Can we ?
> 
> How is that not covered by your upcast mode?

if we generate the adapter-version of asX when X constructor has the adapter, and if class B doesn't have it. 
But this case can well never happen.
What I tried to explain is that ideally we should apply `static_pointer_cast` instead of `static_cast` when the argument of `asX`  contains a shared_ptr. But I don't see how we could make such test.

> 
> > 3. It's armful if the instance of D has an owner which is `!= address` and is not a shared pointer. Since it'll be treated as a shared pointer by the adapter.
> 
> Right, that's going to give a compiler error, so that's not a big issue.

Won't it give a segmentation fault instead ? `owner` will be C-casted from `void*` to `shared_ptr<D>*`.

> > We can check for `explicitUpcast` too to limit the risk, but this will exclude the following case:
> > 
> > * D multiple-inherit from B1 and B2.
> 
> Sure, but that's alright for backward compatibility.
> 
> > * B2 is managed by shared pointers.
> 
> static_pointer_cast doesn't get used unless there is multiple inheritance, so like I keep saying, you'll need to explain a bit more what you're worried about here.
> 
> > That could happen. If we limit to `explicitUpcast`, this case may trigger segmentation faults due to pointers owned by multiple shared pointers.
> 
> Why? You mean we won't be changing the default behavior so that's a bad thing? There has been exactly zero issue reported about that, so there's no reason to fix what isn't broken.

It's not because there was no report that there is nothing to improve. You well see that if someone calls a `asX` on an instance containing a shared pointer, the shared pointer is lost. It's just that no one came into that situation, or on into the one where it can trigger a segmentation fault, or just didn't bother to report and found a workaround, or gave up.

The current behavior will be preserved in all cases since currently no constructors has annotations.

src/main/java/org/bytedeco/javacpp/tools/Parser.java
HGuillemet: In fact `explicitUpcast` should really target a Java name, and not a C++ name. So what can be done is:
* when an info `explicitUpcast` is encountered for a C++ name, we register its associated Java name  in some `Set`.
* we set `type.explicitUpcast` to `true` when the java Name is in the set (or simply get rid of `type.explicitUpcast` and check `set.contains(type.javaName)` instead of checking`type.explicitUpcast`).
Thus no need for a specific code about smart Pointer like above. The fact that C++ `M` and `shared_ptr<M>` are associated with the same Java name is enough.

What do you think about this ?

saudet: Looks good! Now the only place left where strings related to shared_ptr are used is in explicitCast(), so if you could replace that by the actual annotations on the constructor, it should work pretty much like I'm imagining it.

HGuillemet: I changed also this part to just check for the presence of an annotation. I though that's what you suggested. Have you checked the new code ?

saudet: Yes, that's alright, but instead of doing if(constructorIsAnnotated(t)) ... @SharedPtr ... do something like String a = constructorAnnotations(); if (a != null) ... a ...

HGuillemet: Ok. Done.

src/main/java/org/bytedeco/javacpp/tools/Parser.java
```diff
@@ -3233,6 +3302,34 @@ boolean using(Context context, DeclarationList declList) throws ParserException
...
+        if (override) res += "@Override ";
+        res += "public " + to.javaName + ' ' + ecmn + "() { return " + ecmn + "(this); }\n";
+        /* We generate the adapter-aware version of the method when the target class has any annotation,
+           assuming that this annotation is @SharedPtr or some other adapter storing a share_ptr in owner.
+           This is the only use case for now. We can generalize to any adapter if the need arises and call
+           some cast function on the adapter instead of static_pointer_cast. */
+        String annotation = getConstructorAdapter(to);
+        if (annotation != null)
+            res += "    @Namespace public static native " + annotation + " @Name(\"SHARED_PTR_NAMESPACE::static_pointer_cast<" + to.cppName + ">\") "
+                + to.javaName + " " + ecmn + "(@Cast({\"\", \"SHARED_PTR_NAMESPACE::shared_ptr<" + from.cppName + ">\"}) " + annotation + " " + from.javaName + " pointer);\n";
```
saudet, line 3317: If you need a cast here for this to work, put it on the constructor in your presets, not here.

src/main/java/org/bytedeco/javacpp/tools/Context.java
```diff
@@ -53,6 +53,7 @@ class Context {
         templateMap = c.templateMap;
         usingList = c.usingList;
         namespaceMap = c.namespaceMap;
+        explicitUpcast = c.explicitUpcast;
```
saudet, line 56: By definition, "casting" is always *explicit*, so please strip out that term from everywhere.

HGuillemet: What about "static" instead of "explicit" ? Cast alone is used elsewhere else will add confusion for the code reader.

src/main/java/org/bytedeco/javacpp/tools/Parser.java
HGuillemet: I replaced the `@Cast` by making explicit of the second template parameter in `static_pointer_cast`. Commit pushed.

src/main/java/org/bytedeco/javacpp/tools/Context.java
saudet: "upcast" isn't used anywhere else.

src/main/java/org/bytedeco/javacpp/tools/Parser.java
```diff
@@ -3233,6 +3302,34 @@ boolean using(Context context, DeclarationList declList) throws ParserException
         return true;
     }
 
+    private String explicitCast(Type from, Type to, boolean override) {
+        String ecmn = explicitCastMethodName(to.javaName);
```
saudet, line 3306: If you're talking about these functions, you never call them to produce a downcast, so we can rename them to upcast() and upcastMethodName(), right?

HGuillemet: Done.

src/main/java/org/bytedeco/javacpp/annotation/Virtual.java
```diff
@@ -21,4 +21,5 @@
     /** Pure (abstract) or not. */
     boolean value() default false;
     boolean subclasses() default true;
-}
\ No newline at end of file
+    String javaName() default "";
```
saudet, line 24: I wonder if we can have a better name for that field. Maybe "callback"?

src/main/java/org/bytedeco/javacpp/tools/Parser.java
```diff
@@ -938,9 +962,13 @@ Type type(Context context, boolean definition) throws ParserException {
             }
         }
 
-        if (info != null && info.cppTypes != null && info.cppTypes.length > 0) {
+        if (info != null && info.cppTypes != null && info.cppTypes.length > 0 && !info.cppTypes[0].equals(type.cppName)) {
             // use user defined type
             type.cppName = info.cppTypes[0];
```
saudet, line 967: Do we need this? Why did you add this condition in the first place?

src/main/java/org/bytedeco/javacpp/tools/Parser.java
```diff
@@ -2515,9 +2545,22 @@ boolean function(Context context, DeclarationList declList) throws ParserExcepti
                 }
             }
 
+            boolean needWrapper;
```
saudet, line 2548: Could we more meaningfully name that variable something like "needUpcast"?

src/main/java/org/bytedeco/javacpp/tools/Parser.java
HGuillemet: No we don't. 
I added it when I needed the two lines now in comment below, to avoid an unnecessary call to a new Parser.
The two lines in comment are not strictly needed now, since `type.arguments` is not used further (I think).
If you don't want to uncomment them for now, let's remove the extra condition.

src/main/java/org/bytedeco/javacpp/tools/Parser.java
HGuillemet: Ok. Or needUpcastWrapper. It's true if we need a wrapper to upcast `this` and/or at least one argument.

src/main/java/org/bytedeco/javacpp/annotation/Virtual.java
HGuillemet: Why not. Or `methodName` ? Strictly speaking a java method overriding a c++ method is not really a callback (some function reference we pass as an argument). I agree `javaName` is not great.
What about renaming `value` to `pure` BTW ? I doubt it will break anything since no one used the parameter name `value` when it was the single parameter. Does value make some sense ?

saudet: > Why not. Or `methodName` ? Strictly speaking a java method overriding a c++ method is not really a callback (some function reference we pass as an argument). I agree `javaName` is not great.

Right, what about just "method"?

> What about renaming `value` to `pure` BTW ? I doubt it will break anything since no one used the parameter name `value` when it was the single parameter. Does value make some sense ?

The name "value" has a special meaning for annotations, we can't change that without breaking compatibility, no.

HGuillemet: Ah, right.
Commit pushed.

src/main/java/org/bytedeco/javacpp/tools/Parser.java
saudet: Sure, if you really want to have the word "wrapper" in there.

src/main/java/org/bytedeco/javacpp/tools/Parser.java
saudet: Yeah, no, it's alright like that.

saudet:
This looks pretty much good to merge. I've adjusted a few nits, so could you confirm that it still does what you need it to do, without changing the results for any of the presets?

src/main/java/org/bytedeco/javacpp/tools/Parser.java
```diff
@@ -2594,7 +2595,7 @@ boolean function(Context context, DeclarationList declList) throws ParserExcepti
                               +  "_" + dcl.javaName + (dcl.parameters.names == null ? "()" : dcl.parameters.names) + "; }\n";
                     dcl.javaName = "_" + dcl.javaName;
                 }
-                decl.text += modifiers + type.annotations + context.shorten(type.javaName) + " " + dcl.javaName + dcl.parameters.list + ";\n";
+                decl.text += nativeModifiers + type.annotations + context.shorten(type.javaName) + " " + dcl.javaName + dcl.parameters.list + ";\n";
```
HGuillemet, line 2598: We don't want to modifier `modifiers` here because it's reused in next iterations of the `for (int n = -2; n < Integer.MAX_VALUE; n++)` loop.

HGuillemet:
With last commit, still working and no changes in presets parsing compared to master.

src/main/java/org/bytedeco/javacpp/tools/Parser.java
saudet: Right, ok, that `modifiers` though is also "native". Let's name the copy `privateModifiers` maybe if you want to give it a meaningful name? I'd just go with `modifiers2`, but anyway, `privateModifiers` sounds good.

HGuillemet: Done

PR #677, opened 2023-05-12
Fix template substitution with const

HGuillemet:
```c++
template <typename T>
class X {
 public:
    X(const T*) {}
};
```

If you instantiate this template with `int *`, then the resulting type should be `int * const *` and not `const int **`.  
But the parser produces this, with a `new Info("X<int *>").pointerTypes("XI")`:
``` Java
    public XI(@Cast("const int**") PointerPointer arg0) { super((Pointer)null); allocate(arg0); }
    private native void allocate(@Cast("const int**") PointerPointer arg0);
```
Leading to JNI compile error since the constructor expects a `int * const*`.
This PR tries to fix this. It seems to do the trick. The parser now emits:
``` Java
    public XI(@Cast("int**") PointerPointer arg0) { super((Pointer)null); allocate(arg0); }
    private native void allocate(@Cast("int**") PointerPointer arg0);
```
The `const` doesn't appear in the cast at all, but at least we have a better chance to compile the JNI.
I'm not sure it generalizes well. Can we do something better ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 
+ * Fix `Parser` incorrectly mapping `const` pointers to template arguments of pointer types ([pull #677](https://github.com/bytedeco/javacpp/pull/677))
  * Fix `Parser` with `Info.enumerate` failing to translate `enum` values based on other `enum` values
  * Fix `Parser` prematurely expanding macros defined in `class`, `struct` or `union` ([issue #674](https://github.com/bytedeco/javacpp/issues/674))
  * Add `Info.upcast` to support class hierarchies with virtual inheritance ([pull #671](https://github.com/bytedeco/javacpp/pull/671))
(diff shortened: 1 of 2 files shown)


saudet:
Info.constPointer currently gets ignored in this case, yes. It's complicated, but if this fix works for that template that's fine for now. We can do something about Info.constPointer later, I think that's OK, yes. Did you make sure it doesn't break anything in the presets though?

HGuillemet:
Just finished to reparse (almost) all presets. Only changes are for pytorch where `const` is removed from `@Cast` in the following declarations:

In BlockArrayRef.java:
```
public native @Cast("torch::jit::Block**") PointerPointer data();
```

In ValueArrayRef.java
```
public native @Cast("torch::jit::Value**") PointerPointer data();
```

JNI still compiles.

PR #689, opened 2023-06-12
Add properties jniinclude and jnicinclude

HGuillemet:
Add new properties for declaring separately the list of headers to parse and the list of headers to include in JNI code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 
+ * Add `@Platform(jnicinclude=..., jniinclude=...)` properties that `Generator` prioritizes ([pull #689](https://github.com/bytedeco/javacpp/pull/689))
  * Fix `Parser` on function pointer declarations starting with `typedef struct` ([pull bytedeco/javacpp-presets#1361](https://github.com/bytedeco/javacpp-presets/pull/1361))
 
 ### June 6, 2023 version 1.5.9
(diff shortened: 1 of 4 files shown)


saudet:
This looks OK, but I don't think I'm going to merge it because I realized we can already do what you want by modifying the platform.include property inside an `init()` method like this one here: https://github.com/bytedeco/javacpp-presets/blob/master/pytorch/src/main/java/org/bytedeco/pytorch/presets/torch.java#L1795

In your case, I'd put the top-level header files in `@Platform(include=...` and leave it there when `init()` gets called by the Generator, but modify it as required when it gets called by the Parser. We can literally code anything in there, including calling clang from the LLVM presets, so it's a lot more flexible than an annotation value.

HGuillemet:
I considered this option, but currently [`allIncludes` is filled from include properties before init is called](https://github.com/bytedeco/javacpp/blob/b9f91d7f41e99f0f209fbc5ef318ef2217c89dc8/src/main/java/org/bytedeco/javacpp/tools/Parser.java#L4354-L4389). 
And I don't think `init()` is called by Generator. is it ?

saudet:
The init() method gets called as part of Loader.loadProperties(), at the very beginning of Parser.parse(), same for Generator.

HGuillemet:
Ok. I'll load this include list from a resource file from init().
How to know if init is called from parser or from generator ?

saudet:
We might want to add something for that, but just checking the stack with Loader.getCallerClass() or something should suffice, so if that works I'm inclined to use that.

PR #693, opened 2023-07-04
Fix parsing of functions parameters in template arguments

HGuillemet:
This PR proposes to call `parameters` from `templateArguments` when a function is found in order to correctly pass its prototype.

This solves  two issues: the parsing of argument identifiers when present, and the resolution of namespaces.

Example:

```c++
namespace ns {
  struct S {
    int x;
  };

  std::function<void(int a)> a;
  std::function<S(S)> b;
}
```
Before, parsed as:
```
std::function<void(inta)>
std::function<ns::S(S)>
```
with the PR:
```
std::function<void(int)>
std::function<ns::S(ns::S)>
```
Some presets need to be adjusted to parse correctly with this PR (fixing of `std::function` parameters):
* 4 mappings in current pytorch
* 2 in opencv
* 1 in onnx

No changes in other presets.
Not tested: libfreenect2, chilitags,hyperscan, tensorflow, ale, ngraph, qt, skia, videoinput, tritonserver, depthai

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 
+ * Fix `Parser` for function parameters contained in template arguments ([pull #693](https://github.com/bytedeco/javacpp/pull/693))
  * Fix `Parser` on function pointer declarations starting with `typedef struct` ([pull bytedeco/javacpp-presets#1361](https://github.com/bytedeco/javacpp-presets/pull/1361))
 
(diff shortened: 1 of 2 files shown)


HGuillemet:
Do you want a PR on javacpp-presets for the corresponding fixes ?

saudet:
No need to, they'll get fixed when they get updated.

saudet (approved):

PR #700, opened 2023-08-05
Add of for downcasting when c-style cast cannot work

HGuillemet:
The `asB()` method generated by the Parser allows to upcast an instance of a class `D` to an instance of one of its base class `B` in the cases where C-style pointer cast (as used for instance in pointer cast constructors `B(Pointer)` does not work.

This is useful for multiple inheritance and for virtual inheritance from a polymorphic class.

But we lack the equivalent downcast.
```
struct B1 {
  int i;
};
struct B2 {
  int j;
};
struct D: public B1, public B2 {
  int k;
}
```
In Java, `D` extends `B1` and has a `asB2()` method.
But if we have an instance `b2` of `B2` that we know is a `D`, we have no way to cast it to an instance of `D`. `new D(b2)` will lead to a segmentation fault.

For polymorphic classes, here is an example from Pytorch: `LinearImpl` extends `Module` (through `LinearImplCloneable`). Say we get a instance of `Module` from one of the many functions of libtorch returning a Module, we know it's a `LinearImpl`, and we need to call its `forward` function. We cannot. `new LinearImpl(module).forward(tensor)` triggers a segmentation fault (see issue https://github.com/bytedeco/javacpp-presets/issues/1393).

This PR adds a static `of(B)` method, along with the `asB()` method, to perform the downcast. This static method does a `static_cast`/`static_pointer_cast` or a `dynamic_cast`/`dynamic_pointer_cast`. The C++ rules that determine which casting flavor to use are as follows:
* When downcasting through virtual inheritance, static cast cannot be used.
* Dynamic cast can only be used if the source class is polymorphic (having at least 1 virtual member, declared or inherited).

These rules imply that downcasting is not possible (without hacks we won't address) if inheritance is virtual but the base class is not polymorphic.

Virtual inheritance can be reliably detected by the parser.
Polymorphism can be detected too, but not reliably since a class can inherit virtual members from a class defined in a non-parsed, or non-parsed-yet,  header.
This PR also adds these detections to determine if downcast can be performed and which flavor to use. It also adds a `polymorphic` info to label a class as polymorphic when the Parser cannot detect it.
Not that `upcast` info is still needed because we most certainly need this information before we parse the subclass that declares a virtual inheritance and we realize `upcast` is needed. However I added a warning when we parse the subclass if the situation is detected and the `upcast` info has not been set.

Changes in existing presets: addition of `of` in presets with classes having multiple or virtual inheritance (hdf5, opencv, pytorch and tvm)

COULD BE IMPROVED:
It would be better if `of` were replaced by a constructor, acting as a specialization of the pointer cast constructor taking a `Pointer`. This way we limit the risk that the user triggers a segmentation fault by calling the pointer cast constructor when C-style cast won't work. But I don't see how to implement this easily as a constructor, without patching the generator. We could add something like:
```
public D(B pointer) { super(of(B)); }
```
and make `of` private, but it's not really satisfactory since it just duplicates the object returned by `of`. Any idea ?

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/annotation/Adapter.java
+++ b/src/main/java/org/bytedeco/javacpp/annotation/Adapter.java
@@ -10,22 +10,82 @@ import org.bytedeco.javacpp.Pointer;
 import org.bytedeco.javacpp.tools.Generator;
 
 /**
- * Specifies a C++ class to act as an adapter to convert the types of arguments.
- * Three such C++ classes made available by {@link Generator} are {@code StringAdapter},
- * {@code VectorAdapter}, and {@code SharedPtrAdapter} to bridge a few differences between
- * {@code std::string} and {@link String}; between {@code std::vector}, Java arrays of
- * primitive types, {@link Buffer}, and {@link Pointer}; and between {@code xyz::shared_ptr}
- * and {@link Pointer}. Adapter classes must define the following public members:
- * <ul>
- * <li> A constructor accepting 3 arguments (or more if {@link #argc()} > 1): a pointer, a size, and the owner pointer
- * <li> Another constructor that accepts a reference to the object of the other class
- * <li> A {@code static void deallocate(owner)} function
- * <li> Overloaded cast operators to both types, for references and pointers
- * <li> A {@code void assign(pointer, size, owner)} function
- * <li> A {@code size} member variable for arrays accessed via pointer
- * </ul>
- * To reduce further the amount of coding, this annotation can also be used on
- * other annotations, such as with {@link StdString}, {@link StdVector}, and {@link SharedPtr}.
+ * Specifies a C++ class to act as an adapter between a target type and one or more adaptee type(s).
+ * Instances of the adapter class are short-living and last only for the duration of a JNI call.
+ * <p></p>
+ * Six such C++ classes are made available by {@link Generator}:
+ * <blockquote>
+ * <table width="80%">
+ *     <thead>
+ *         <tr>
+ *             <th>Adapter class</th><th>Target type</th><th>Adaptee types</th><th>Helper annotation</th>
+ *         </tr>
+ *     </thead>
+ *     <tbody>
+ *         <tr>
+ *             <td valign="top">{@code VectorAdapter<P,T,A>}</td>
+ *             <td valign="top">{@code std::vector<T,A>}</td>
+ *             <td valign="top">{@code P}</td>
+ *             <td valign="top">{@link StdVector}</td>
+ *         </tr>
+ *         <tr>
+ *             <td valign="top">{@code StringAdapter<T>}</td>
+ *             <td valign="top">{@code std::basic_string<T>}</td>
+ *             <td valign="top">{@code char}<br>{@code signed char}<br>{@code unsigned char}<br>{@code wchar_t}<br>{@code unsigned short}<br>{@code signed int}</td>
+ *             <td valign="top">{@link StdString}</td>
+ *         </tr>
+ *         <tr>
+ *             <td valign="top">{@code SharedPtrAdapter<T>}</td>
+ *             <td valign="top">{@code SHARED_PTR_NAMESPACE::shared_ptr<T>}</td>
+ *             <td valign="top">{@code T}</td>
+ *             <td valign="top">{@link SharedPtr}</td>
+ *         </tr>
+ *         <tr>
+ *             <td valign="top">{@code UniquePtrAdapter<T,D>}</td>
+ *             <td valign="top">{@code UNIQUE_PTR_NAMESPACE::unique_ptr<T,D>}</td>
+ *             <td valign="top">{@code T}</td>
+ *             <td valign="top">{@link UniquePtr}</td>
+ *         </tr>
+ *         <tr>
+ *             <td valign="top">{@code MoveAdapter<T,D>}</td>
+ *             <td valign="top">{@code T}</td>
+ *             <td valign="top">{@code T}</td>
+ *             <td valign="top">{@link StdMove}</td>
+ *         </tr>
+ *         <tr>
+ *             <td valign="top">{@code OptionalAdapter<T>}</td>
+ *             <td valign="top">{@code OPTIONAL_NAMESPACE::optional<T>}</td>
+ *             <td valign="top">{@code T}</td>
+ *             <td valign="top">{@link Optional}</td>
+ *         </tr>
+ *     </tbody>
+ * </table>
+ * </blockquote>
+ * The helper annotations are shortcuts that infer the template type(s) of the adapter class from the Java
+ * class they annotate.
+ * <p></p>
+ * When an argument of a method is annotated, an instance of the adapter class is created from
+ * the Java object passed as argument, and this instance is passed to the C++ function, thus triggering
+ * an implicit cast to the type expected by the function (usually a reference or pointer to the target type).
+ * If the argument is also annotated with {@link Cast},
+ * the adapter instance is cast to the type(s) specified by the {@link Cast} annotation before being passed
+ * to the function.
+ * <p></p>
+ * When a method is annotated, an instance of the adapter is created from the
+ * value (usually a pointer or reference to the target type) returned by the C++ function or by {@code new} if the method is an allocator.
+ * If the method is also annotated with {@link Cast}, the value returned by the C++ function is
+ * cast by value 3 of the {@link Cast} annotation, if any, before instantiation of the adapter.
+ * Then a Java object is created from the adapter to be returned by the method.
+ * <p></p>
+ * Adapter classes must at least define the following public members:
+ *  <ul>
+ *  <li> For each adaptee type, a constructor accepting 3 arguments (or more if {@link #argc()} > 1): a pointer to a const value of the adaptee, a size, and the owner pointer
+ *  <li> Another constructor that accepts a reference to the target type
+ *  <li> A {@code static void deallocate(owner)} function
+ *  <li> Overloaded cast operators to both the target type and the adaptee types, for references and pointers
+ *  <li> {@code void assign(pointer, size, owner)} functions with the same signature than the constructors accepting 3 arguments
+ *  <li> A {@code size} member variable for arrays accessed via pointer
+ *  </ul>
  *
  * @see Generator
  *
--- a/src/main/java/org/bytedeco/javacpp/annotation/Adapter.java
+++ b/src/main/java/org/bytedeco/javacpp/annotation/Adapter.java
@@ -14,8 +14,8 @@ import org.bytedeco.javacpp.tools.Generator;
  * Instances of the adapter class are short-living and last only for the duration of a JNI call.
  * <p></p>
  * Six such C++ classes are made available by {@link Generator} to bridge a few differences, for instance,
- * between {@code std::string} and {@link String}, between {@code std::vector}, Java arrays of
- *  primitive types, {@link Buffer}, and {@link Pointer}, or between {@code xyz::shared_ptr} and {@link Pointer}:
+ * between {@code std::string} and {@link String}, between {@code std::vector}, Java arrays of primitive
+ * types, {@link Buffer}, and {@link Pointer}, or between {@code xyz::shared_ptr} and {@link Pointer}:
  * <blockquote>
  * <table width="80%">
  *     <thead>
@@ -69,12 +69,11 @@ import org.bytedeco.javacpp.tools.Generator;
  * When an argument of a method is annotated, an instance of the adapter class is created from
  * the Java object passed as argument, and this instance is passed to the C++ function, thus triggering
  * an implicit cast to the type expected by the function (usually a reference or pointer to the target type).
- * If the argument is also annotated with {@link Cast},
- * the adapter instance is cast to the type(s) specified by the {@link Cast} annotation before being passed
- * to the function.
+ * If the argument is also annotated with {@link Cast}, the adapter instance is cast to the type(s) specified
+ * by the {@link Cast} annotation before being passed to the function.
  * <p></p>
- * When a method is annotated, an instance of the adapter is created from the
- * value (usually a pointer or reference to the target type) returned by the C++ function or by {@code new} if the method is an allocator.
+ * When a method is annotated, an instance of the adapter is created from the value (usually a pointer or
+ * reference to the target type) returned by the C++ function or by {@code new} if the method is an allocator.
  * If the method is also annotated with {@link Cast}, the value returned by the C++ function is
  * cast by value 3 of the {@link Cast} annotation, if any, before instantiation of the adapter.
  * Then a Java object is created from the adapter to be returned by the method.
@@ -88,6 +87,8 @@ import org.bytedeco.javacpp.tools.Generator;
  *  <li> {@code void assign(pointer, size, owner)} functions with the same signature than the constructors accepting 3 arguments
  *  <li> A {@code size} member variable for arrays accessed via pointer
  *  </ul>
+ * To reduce further the amount of coding, this annotation can also be used on
+ * other annotations, such as with {@link StdString}, {@link StdVector}, and {@link SharedPtr}.
  *
  * @see Generator
  *
--- a/src/main/java/org/bytedeco/javacpp/annotation/Adapter.java
+++ b/src/main/java/org/bytedeco/javacpp/annotation/Adapter.java
@@ -13,7 +13,9 @@ import org.bytedeco.javacpp.tools.Generator;
  * Specifies a C++ class to act as an adapter between a target type and one or more adaptee type(s).
  * Instances of the adapter class are short-living and last only for the duration of a JNI call.
  * <p></p>
- * Six such C++ classes are made available by {@link Generator}:
+ * Six such C++ classes are made available by {@link Generator} to bridge a few differences, for instance,
+ * between {@code std::string} and {@link String}, between {@code std::vector}, Java arrays of
+ *  primitive types, {@link Buffer}, and {@link Pointer}, or between {@code xyz::shared_ptr} and {@link Pointer}:
  * <blockquote>
  * <table width="80%">
  *     <thead>
--- a/src/main/java/org/bytedeco/javacpp/tools/Generator.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Generator.java
@@ -2645,11 +2645,16 @@ public class Generator {
                     prefix = "";
                     suffix = "";
                 } else {
-                    out.print((noException(methodInfo.cls, methodInfo.method) ?
-                        "new (std::nothrow) " : "new ") + valueTypeName + typeName[1]);
-                    if (methodInfo.arrayAllocator) {
-                        prefix = "[";
-                        suffix = "]";
+                    if (methodInfo.method.isAnnotationPresent(Name.class)) {
+                        out.print(methodInfo.memberName[0]);
+                        // If method is an array allocator, the function must return a pointer to an array
+                    } else {
+                        out.print((noException(methodInfo.cls, methodInfo.method) ?
+                            "new (std::nothrow) " : "new ") + valueTypeName + typeName[1]);
+                        if (methodInfo.arrayAllocator) {
+                            prefix = "[";
+                            suffix = "]";
+                        }
                     }
                 }
             } else if (Modifier.isStatic(methodInfo.modifiers) || !Pointer.class.isAssignableFrom(methodInfo.cls)) {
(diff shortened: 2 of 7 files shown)


saudet:
I'm not entirely sure I understand what you're trying to do here, but it's obviously more complicated than it needs to be. I'll let you go through a few iterations of this until you settle on something that makes more sense.

HGuillemet:
I was probably not clear enough. What don't you understand in the purpose of this PR ?
How would you downcast a pytorch `Module` to `LinearImpl` in a simpler way ?
Basically, this PR just introduces the inverse of the `asXXX` methods that perform an upcast in the case of multiple inheritance and virtual inheritance.

saudet:
Among probably other things, I think we can get away with only dynamic_cast, and if we do that, how much does it simplify?

saudet:
I see, so there's too many types of casts to try and force this into never ending flags. How about we add an Info that would let users specify the name of the cast function (ehem, operator) they want to use for this?

HGuillemet:
The PR implements the needed logic to make the right choice between static_cast and dynamic_cast automatically, without needing info (but the new polymorphic info, added for good measure, but which may well be never used).
It was tested in the different scenarios: polymorphic base or not, virtual inheritance or not.

Using an explicit info as you suggest would work, but why request that from the user when we the parser can detect what to do by itself ? To save about ~ 10 new lines of code in parser ?
Also this new lines of code allows, as a side-effect, to emit a warning when `upcast` info has not been provided when it should be.

saudet:
So let's remove that flag entirely until it's necessary somehow for someone somewhere?

HGuillemet:
polymorphic ? Ok, as you wish.

saudet:
But this behavior shouldn't be enabled by default. It only happens on Info.upcast right? That's a bit weird..

Anyway, from() sounds better than of() for that kind of operation:
https://docs.oracle.com/javase/tutorial/datetime/overview/naming.html

HGuillemet:
> But this behavior shouldn't be enabled by default. It only happens on Info.upcast right? That's a bit weird..

Which behavior ? The generation of the of/from methods ?
They are generated when asX methods are, so either with virtual inheritance (`upcast` case), or multiple inheritance (automatic).
The method name `upcast` in Parser should be renamed, as `upAndDownCast` or whatever name you prefer.


> 
> Anyway, from() sounds better than of() for that kind of operation: https://docs.oracle.com/javase/tutorial/datetime/overview/naming.html

You must be right. However I think the best would be a constructor, as said in the last note or the top post. Any opinion or idea about that ?

HGuillemet:
> But this behavior shouldn't be enabled by default. It only happens on Info.upcast right? That's a bit weird..

The `upcast` name for the info turns out to be not a good choice either, with this new feature. `polymorphicBaseClassWithDerivedClassesInheritingFromIt` probably not either. Any better idea ?

saudet:
> You must be right. However I think the best would be a constructor, as said in the last note or the top post. Any opinion or idea about that ?

I suppose you could add new constructors like you proposed? It creates a temporary wrapper object, but it can't hurt.

> The `upcast` name for the info turns out to be not a good choice either, with this new feature. `polymorphicBaseClassWithDerivedClassesInheritingFromIt` probably not either. Any better idea ?

Meh, Info.upcast is fine, let's not add anything more just for this. Ideally we'd want to always do all of that upcasting and downcasting I suppose, but for backward compatibility, let's just make sure upcast() doesn't get called without Info.upcast.

HGuillemet:
> > You must be right. However I think the best would be a constructor, as said in the last note or the top post. Any opinion or idea about that ?
> 
> I suppose you could add new constructors like you proposed? It creates a temporary wrapper object, but it can't hurt.

Ok. I guess we could add support for `@Name` on allocate methods to directly do what is done by of/from in a constructor. Can you think of other uses that would make this addition to the generator worth ? Else let's stick with the wrapper, at least for now.

> > The `upcast` name for the info turns out to be not a good choice either, with this new feature. `polymorphicBaseClassWithDerivedClassesInheritingFromIt` probably not either. Any better idea ?
> 
> Meh, Info.upcast is fine, let's not add anything more just for this. Ideally we'd want to always do all of that upcasting and downcasting I suppose, but for backward compatibility, let's just make sure upcast() doesn't get called without Info.upcast.

The current upcast() is already called both for multiple inheritance and `upcast`. It generates methods for upcasting only.
The PR adds the generation of reciprocal downcasting methods to it.
Do you mean that you prefer to generate the new downcast constructor for virtual inheritance/`upcast` only and not for multiple inheritance ?
That seems weird for me, since that won't simplify the code. And the PR won't break anything, it fixes a "bug" triggered when calling the pointer cast constructor and causing a segmentation fault (even if nobody reported it yet).

saudet:
> Ok. I guess we could add support for `@Name` on allocate methods to directly do what is done by of/from in a constructor. Can you think of other uses that would make this addition to the generator worth ? Else let's stick with the wrapper, at least for now.

Sounds reasonable, `@Name` isn't used for anything on allocate() methods right now. That could be used to call random create functions, to "objectify" C APIs and what not, as well.

> The current upcast() is already called both for multiple inheritance and `upcast`. It generates methods for upcasting only. The PR adds the generation of reciprocal downcasting methods to it. Do you mean that you prefer to generate the new downcast constructor for virtual inheritance/`upcast` only and not for multiple inheritance ? That seems weird for me, since that won't simplify the code. And the PR won't break anything, it fixes a "bug" triggered when calling the pointer cast constructor and causing a segmentation fault (even if nobody reported it yet).

No, I mean it shouldn't generate anything, unless the class has the Info.upcast flag explicitly set by the user.

HGuillemet:
> Sounds reasonable, `@Name` isn't used for anything on allocate() methods right now. That could be used to call random create functions, to "objectify" C APIs and what not, as well.

Do you prefer I look into this and try to make a PR, or will you do it yourself ?

> 
> > The current upcast() is already called both for multiple inheritance and `upcast`. It generates methods for upcasting only. The PR adds the generation of reciprocal downcasting methods to it. Do you mean that you prefer to generate the new downcast constructor for virtual inheritance/`upcast` only and not for multiple inheritance ? That seems weird for me, since that won't simplify the code. And the PR won't break anything, it fixes a "bug" triggered when calling the pointer cast constructor and causing a segmentation fault (even if nobody reported it yet).
> 
> No, I mean it shouldn't generate anything, unless the class has the Info.upcast flag explicitly set by the user.

I don't get it.
Classes with multiple inheritance already have the `asB2()` method for upcasting to the class they don't inherit from in Java. This was the case before the addition of the `upcast` info. 
You propose to add the downcast constructor `D(B2 b2)` only if we set Info.upcast on B2 ? This is weird.
Also I remind you that the upcast info does add the generation of `asB()` upcast method, but it's main interest is to add wrappers to methods taking a B as argument so that we can pass a D without triggering a segmentation fault. This is for virtual inheritance only doesn't concern multiple inheritance.

saudet:
> Do you prefer I look into this and try to make a PR, or will you do it yourself ?

It shouldn't be too hard, but it's not something I'll be doing myself, no...

> I don't get it. Classes with multiple inheritance already have the `asB2()` method for upcasting to the class they don't inherit from in Java. This was the case before the addition of the `upcast` info. You propose to add the downcast constructor `D(B2 b2)` only if we set Info.upcast on B2 ? This is weird. Also I remind you that the upcast info does add the generation of `asB()` upcast method, but it's main interest is to add wrappers to methods taking a B as argument so that we can pass a D without triggering a segmentation fault. This is for virtual inheritance only and doesn't concern multiple inheritance.

Right, ok, so let's say, if doesn't break anything with the presets, we can leave like that, sounds good?

HGuillemet:
> Right, ok, so let's say, if doesn't break anything with the presets, we can leave like that, sounds good?

Leave the PR as is ? Sure.
But let me have a look and see if I manage to add @Name on allocate.
And we can rename method upcast (not info upcast), but that's a detail.

saudet:
> Leave the PR as is ? Sure. But let me have a look and see if I manage to add `@Name` on allocate. And we can rename method upcast (not info upcast), but that's a detail.

We probably only need to update this line:
https://github.com/bytedeco/javacpp/blob/1.5.9/src/main/java/org/bytedeco/javacpp/tools/Generator.java#L2648

HGuillemet:
Right. That's what I did.

However, there is a little complication. Could you advise about it ?

Currently, when `allocate` is annoted with an adapter, like `@SharedPtr`, it constructs an adapter built with:
```
adapter(ptr, 1, NULL)
```
where `ptr` results from a `new`.

Here we want, in addition, to annotate with something like `@Name("dynamic_pointer_cast<...>")` to replace the `new`, so `ptr` is a `shared_ptr`. 
We thus need to call the 1-argument  constructor of the adapter. 

What would be the best way to inform the generator about this ?

saudet:
I'm not sure I follow, I didn't write that code. You're the one that added
support for that, so do whatever it needs to do?

src/main/java/org/bytedeco/javacpp/tools/Generator.java
```diff
@@ -2814,12 +2819,8 @@ void returnAfter(MethodInformation methodInfo) {
             // special considerations for std::string without adapter
             out.print(");\n" + indent + "rptr = rstr.c_str()");
         }
-        if (adapterInfo != null) {
-            if (methodInfo.allocator || methodInfo.arrayAllocator) {
-                suffix = ", 1, NULL)" + suffix;
-            } else if (!methodInfo.returnType.isPrimitive()) {
-                suffix = ")" + suffix;
-            }
+        if ((methodInfo.allocator || methodInfo.arrayAllocator || !methodInfo.returnType.isPrimitive()) && adapterInfo != null) {
```
saudet, line 2822: Since the returnType of allocator is void, we don't need to check explicitly and can revert completely to 
```suggestion
        if (!methodInfo.returnType.isPrimitive() && adapterInfo != null) {
```

src/main/java/org/bytedeco/javacpp/tools/Parser.java
```diff
@@ -96,7 +97,17 @@ public Parser(Logger logger, Properties properties, String encoding, String line
     InfoMap leafInfoMap = null;
     TokenIndexer tokens = null;
     String lineSeparator = null;
-    Set<String> upcasts = new HashSet<>();
+    Set<String> upcasts = new HashSet<>(); // Java names of base classes needing upcast
+    Map<String, Set<Type>> downcasts = new HashMap<>(); // Keys: classes which child classes needs downcast constructor. Values: classes to downcast from.
+    HashSet<String> polymorphicClasses = new HashSet<>(); // Contains Java names
```
saudet, line 102: Declare that as `Set` instead of `HashSet` for consistency, and while we're at it, use Javadoc-style comments

src/main/java/org/bytedeco/javacpp/tools/Parser.java
```diff
@@ -3774,18 +3865,22 @@ boolean group(Context context, DeclarationList declList) throws ParserException
...
+            if (addArrayConstructor) {
+                constructors += "    /** Native array allocator. Access with {@link Pointer#position(long)}. */\n" +
+                                "    public " + shortName + "(long size) { super((Pointer)null); allocateArray(size); }\n" +
+                                "    private native void allocateArray(long size);\n" +
+                                "    @Override public " + shortName + " position(long position) {\n" +
+                                "        return (" + shortName + ")super.position(position);\n" +
+                                "    }\n" +
+                                "    @Override public " + shortName + " getPointer(long i) {\n" +
+                                "        return new " + shortName + "((Pointer)this).offsetAddress(i);\n" +
+                                "    }\n";
```
saudet, line 3881: Please do that in a way that doesn't change the output for the presets.

src/main/java/org/bytedeco/javacpp/tools/Generator.java
HGuillemet: These is a `!` before `methodInfo.returnType.isPrimitive()`. I believe this check is here because in `returnBefore` we ignore the adapter if the type is a non-void primitive.

src/main/java/org/bytedeco/javacpp/tools/Parser.java
HGuillemet: I rearranged so that the result is closer to what master outputs. But there is still a difference because in master the array constructor and pointer cast constructor are not emitted in the same order whether `implicitConstructor` is true or not.
Do you prefer to keep this version or a version that reproduces exactly the master output but with a uglier code with repetitions ?

src/main/java/org/bytedeco/javacpp/tools/Generator.java
saudet: Ah, void.class.isPrimitive() returns true, interesting...

src/main/java/org/bytedeco/javacpp/tools/Parser.java
saudet: We can put the comments on a single line still, let's do that

src/main/java/org/bytedeco/javacpp/tools/Parser.java
saudet: We don't need repetitions, just declare a like `String arrayConstructor = ...` instead of a boolean flag and use that in your modifications.

HGuillemet: In master, array constructor  and `allocateArray` are interleaved with other lines when `implicitConstructor` is true, so I don't think it would help.
I reverted to master and just added the tests for `constructorAnnotations.isEmpty()`. The parser output is now the same as for master when there are no annotations on constructors. If you think of a nicer way to implement this, feel free to modify yourself.

saudet: Seems a lot better, but could you revert unnecessary changes in formatting?

HGuillemet: I guessed you mean the alignment of + with = in concatenations. Good enough now ?

src/main/java/org/bytedeco/javacpp/tools/Parser.java
```diff
@@ -253,7 +277,7 @@ void containers(Context context, DeclarationList declList) throws ParserExceptio
                     dcl.definition.text = "\n" + dcl.definition.text;
                     decl.declarator = dcl;
                 }
-                LinkedHashSet<Type> typeSet = new LinkedHashSet<>();
+                LinkedList<Type> typeSet = new LinkedList<>();
                 typeSet.addAll(Arrays.asList(firstType, secondType, indexType, valueType));
                 typeSet.addAll(Arrays.asList(containerType.arguments));
```
saudet, line 282: Please revert that, as well as the unnecessary import. That's intentional to prevent duplicates. Please test your changes against all the presets!

src/main/java/org/bytedeco/javacpp/annotation/Adapter.java
```diff
@@ -10,22 +10,82 @@
...
- * <ul>
- * <li> A constructor accepting 3 arguments (or more if {@link #argc()} > 1): a pointer, a size, and the owner pointer
- * <li> Another constructor that accepts a reference to the object of the other class
- * <li> A {@code static void deallocate(owner)} function
- * <li> Overloaded cast operators to both types, for references and pointers
- * <li> A {@code void assign(pointer, size, owner)} function
- * <li> A {@code size} member variable for arrays accessed via pointer
- * </ul>
- * To reduce further the amount of coding, this annotation can also be used on
- * other annotations, such as with {@link StdString}, {@link StdVector}, and {@link SharedPtr}.
```
saudet, line 28: You've added more comments, but you've also removed information. Please add it back

src/main/java/org/bytedeco/javacpp/tools/Parser.java
HGuillemet: This is related to the addition of `hashCode` in `Type`. Before, there was no `hashCode`, so the `LinkedHashSet` was functionally more similar to a `LinkedList`. I added a `hashCode`, consistent with existing `equals`, because I needed a `Set<Type>` elsewhere. So this change in fact keeps the previous behavior.
I did test against all the presets, and in effect without this change I had parsing errors or changes. I didn't investigate further why `typeSet` must not de-duplicate types. We can do it if you want.

src/main/java/org/bytedeco/javacpp/annotation/Adapter.java
HGuillemet: Do you mean the "to bridge difference..." part ? . Better now ?

saudet: No, but just don't delete it is fine, no?

src/main/java/org/bytedeco/javacpp/tools/Parser.java
HGuillemet: Without the hashCode, the effect of using a Set was to remove "exact" duplicates. That is, if `t1 == t2`, it keeps only one. With the hashCode, it removes duplicates that are the same in terms of `equal`, that is that have the same cppName and javaName.
When, for instance, `firstType` and `secondType` are 2 distinct instances of `Type` but with the same cppName, we need to process each of them in the following loop to add the appropriate `annotation`. That's why it didn't work anymore after the addition of `hashCode`.
After the change from a `Set` to a `List`, if `firstType` and `secondType` are the same instance, the instance is processed twice, but that's not a big deal since, if I'm not wrong, there are tests in the body of the loop that prevents the annotations to be changed twice.

src/main/java/org/bytedeco/javacpp/annotation/Adapter.java
HGuillemet: I don't follow you, can you directly change the way you prefer ?
I added back "to bridge..." altering it a bit because the previous version omits adapters that have been added since.
If it's about the "to reduce further coding...", I think the "The helper annotations are shortcuts ...." gives the same information while being a bit more precise, but change as you like.

src/main/java/org/bytedeco/javacpp/tools/Parser.java
saudet: There's no reason to use LinkedList though, so let's use ArrayList, and let's do the same for the other LinkedHashSet

HGuillemet: > There's no reason to use LinkedList though, so let's use ArrayList,

Or a plain array...
Done.

> and let's do the same for the other LinkedHashSet

`infoSet` ?
I'm not sure. are you ?
The `addAll` could result in the addition of the same `Info` several times, and executing the iterations of the loop several times for the same Info seems to have real effects.

src/main/java/org/bytedeco/javacpp/tools/Parser.java
```diff
@@ -3569,9 +3652,21 @@ boolean group(Context context, DeclarationList declList) throws ParserException
...
+        } else if (polymorphicClasses.contains(base.javaName) && base.virtual) {
+            // In this case we know we need upcast
+            casts += upcast(type, base, false);
+            addDowncast(base.cppName, base);
         }
 
+        Set<Type> froms = downcasts.get(base.cppName);
+        if (froms != null)
+            for (Type t: froms) {
+                casts += downcast(type, t);
```
saudet, line 3667: Since this is actually a constructor, let's append to `constructors` instead?

src/main/java/org/bytedeco/javacpp/tools/Parser.java
```diff
@@ -3360,27 +3383,58 @@ boolean using(Context context, DeclarationList declList) throws ParserException
...
+        }
+        String res = "";
+        res += "    /** Downcast constructor. */\n" +
+               "    public " + derived.javaName + "(" + base.javaName + " pointer) { super((Pointer)null); allocate(pointer); }\n";
+        if (annotations.isEmpty()) {
+            res += "    @Namespace private native @Name(\"" + downcastType + "_cast<" + derived.cppName + "*>\") void allocate(" + base.javaName + " pointer);";
+        } else {
+            res += "    @Namespace private native " + annotations + "@Name(\"SHARED_PTR_NAMESPACE::" + downcastType + "_pointer_cast<" + derived.cppName + ", " + base.cppName + ">\") void allocate(" + annotations + base.javaName + " pointer);";
+        }
+        return res;
```
saudet, line 3413: I think there's a missing "\n" here

HGuillemet: Right.

src/main/java/org/bytedeco/javacpp/tools/Parser.java
HGuillemet: Ok, done.
I also added a check in case a constructor with the same signature has not been already added while parsing.

saudet (approved):
This looks good to merge! Anything else you'd like to change before that though?

PR #704, opened 2023-08-14
Support for LoongArch64

Panxuefeng-loongson:


Panxuefeng-loongson:
@saudet RFR

saudet:
Or is this just setting the LP64 flag explicitly, which is already the default anyway on pretty much all Linux distributions? And if so, how is this different from the existent linux-arm64 platform properties?

Panxuefeng-loongson:
> 
Hi,
Thanks for the reply. Debian12 starts official support for LoongArch: https://lists.debian.org/debian-devel-announce/2023/08/msg00000.html. https://cfarm.tetaneutral.net/machines/list/. There are LoongArch machines available for verification on GCC Farm: https://cfarm.tetaneutral.net/machines/list/

Panxuefeng-loongson:
> Or is this just setting the LP64 flag explicitly, which is already the default anyway on pretty much all Linux distributions? And if so, how is this different from the existent linux-arm64 platform properties?

Doesn't seem to make a difference.

saudet:
Ooh, I see, this isn't ARM...

src/main/java/org/bytedeco/javacpp/Loader.java
```diff
@@ -126,6 +126,8 @@ public static String getPlatform() {
                 osArch = "x86_64";
             } else if (osArch.startsWith("aarch64") || osArch.startsWith("armv8") || osArch.startsWith("arm64")) {
                 osArch = "arm64";
+            } else if (osArch.startsWith("loongarch64")) {
+                osArch = "loongarch64";
```
saudet, line 130: This doesn't change the value of osArch, so we don't need to add anything here. Please remove this.

Panxuefeng-loongson: > This doesn't change the value of osArch, so we don't need to add anything here. Please remove this.

Done

saudet: I'm assuming osArch is always "loongarch64" here though. Let me know if this isn't the case.

Panxuefeng-loongson: > I'm assuming osArch is always "loongarch64" here though. Let me know if this isn't the case.
osArch value is always "loongarch64", you've made the right suggestion.
Thanks

Panxuefeng-loongson: > I'm assuming osArch is always "loongarch64" here though. Let me know if this isn't the case.

osArch value is always "loongarch64", you've made the right suggestion.
Thanks

PR #706, opened 2023-08-28
Add support for `@Name` on allocators

HGuillemet:
Add support for `@Name` on `allocate()` and `allocateArray()`.
The functions passed as parameter to `@Name` will replace the call to `new`.
 
This will be used for instance in (see PR #668):
```
@SharedPtr @Name("std::make_shared<xxx>") private native void allocate();
```
or (see PR #700):
```
@Name("std::dynamic_cast<DerivedClass*>") private native void allocate(BaseClass pointer);
```

Note 1: if placed on a array allocator, the function is supposed to return a pointer to an array. For instance resulting from `std::make_shared<xxx>(n)` (unfortunately supported since C++20 only). 

Note 2: in the special case of classes with `@Virtual` methods, the class to instantiate is the proxy class, with its mangled name, so we need to write things like:
```
@SharedPtr @Name("std::make_shared<JavaCPP_torch_0003a_0003ann_0003a_0003aModule>") private native void allocate();
```
which is a bit hacky, but works.

Also included in the PR is an updated and more detailed javadoc comment for `@Adapter`.

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/tools/Generator.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Generator.java
@@ -2819,7 +2819,7 @@ public class Generator {
             // special considerations for std::string without adapter
             out.print(");\n" + indent + "rptr = rstr.c_str()");
         }
-        if (!methodInfo.returnType.isPrimitive() && adapterInfo != null) {
+        if ((methodInfo.allocator || methodInfo.arrayAllocator || !methodInfo.returnType.isPrimitive()) && adapterInfo != null) {
             suffix = ")" + suffix;
         }
         if ((Pointer.class.isAssignableFrom(methodInfo.returnType) ||
--- a/src/main/java/org/bytedeco/javacpp/tools/Generator.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Generator.java
@@ -2645,11 +2645,16 @@ public class Generator {
                     prefix = "";
                     suffix = "";
                 } else {
-                    out.print((noException(methodInfo.cls, methodInfo.method) ?
-                        "new (std::nothrow) " : "new ") + valueTypeName + typeName[1]);
-                    if (methodInfo.arrayAllocator) {
-                        prefix = "[";
-                        suffix = "]";
+                    if (methodInfo.method.isAnnotationPresent(Name.class)) {
+                        out.print(methodInfo.memberName[0]);
+                        // If method is an array allocator, the function must return a pointer to an array
+                    } else {
+                        out.print((noException(methodInfo.cls, methodInfo.method) ?
+                            "new (std::nothrow) " : "new ") + valueTypeName + typeName[1]);
+                        if (methodInfo.arrayAllocator) {
+                            prefix = "[";
+                            suffix = "]";
+                        }
                     }
                 }
             } else if (Modifier.isStatic(methodInfo.modifiers) || !Pointer.class.isAssignableFrom(methodInfo.cls)) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/bytedeco/javacpp/tools/Generator.java
```diff
@@ -2814,12 +2819,8 @@ void returnAfter(MethodInformation methodInfo) {
             // special considerations for std::string without adapter
             out.print(");\n" + indent + "rptr = rstr.c_str()");
         }
-        if (adapterInfo != null) {
-            if (methodInfo.allocator || methodInfo.arrayAllocator) {
-                suffix = ", 1, NULL)" + suffix;
```
saudet, line 2819: Why do we need to remove this? I don't see how that's not going to break anything

HGuillemet: That's a reversal to pre-PR#668.
The only thing it'll break is the new `@SharedPtr` annotation on constructor, that will now need the `@Name` annotation too. But I'm pretty sure only the pytorch presets uses that for now.

HGuillemet: More explanations: `@SharedPtr` on a method means: the native function returns a `shared_ptr`, construct a Java object containing the `shared_ptr` in owner, by instantiating an adapter from this `shared_ptr`.

But for PR #668 I had to change the logic since `new`, called by `allocate` returned a plain pointer. So the meaning of `@SharedPtr` on allocators was different: the native function returns a plain pointer, turn it into a shared pointer,, by instantiating an adapter from the plain pointer (thus the ", 1, NULL)" additional arguments), and build a java object containing it.

This worked, but in addition to alter the logic and requiring these lines in generator to implement this exception to the logic, it prevented to use `@SharedPtr @Name("std::dynamic_pointer_cast<DerivedClass*>")`, since `dynamic_pointer_cast` returns a shared pointer, not a plain pointer.

So I suggest to revert to the common logic where `@SharedPtr` always mean that the native function return a `shared_ptr` and explictly build the `shared_ptr` by replacing `new` by `make_shared` using `@Name`, instead of delegating this to the adapter.

saudet: Ok, sounds alright, but let's try not to change things back and forth like that all the time. I'm assuming you might want to change that a bit further once you start testing this with pull #700, so could you just put all that there and start testing everything together?

HGuillemet: I already tested both PR on pytorch and it's working well. Do you prefer that I merge both PR into #700 and close this one ? Or that I rebase #700 on this branch ?

HGuillemet: One thing that we might want to add before merging is support for array allocator.

About `@Name` (this PR): see Note 1 above. I'm not 100% sure this is the right choice. An alternative would be to make a loop and call the function for each element. We should then allocate with malloc instead of C++ array allocator to prevent the call to the constructors. Any opinion ?

About adapters: I think JNI just doesn't compile if placed on `allocateArray`.

Anyway, currently, when we add an annotation on constructor with an `Info`, Parser doesn't place it on `allocateArray`. But that is close to a bug, since for instance, if someone ever calls:
```
ReLUImpl relus = new ReLUImpl(10);
```
Each ReLUImpl will not have been created with `make_shared`, Potentially leading to memory corruption.

HGuillemet: Also see Note 2. A possible improvement is to find a way to use a predictable name for proxy classes of virtualized class. Any idea about this ?

HGuillemet: Ok, I'll merge this into #700. 

Concerning arrayAllocators: do you agree with the fact that shared pointer adapter just cannot work with arrays of objects or did I miss something ? In this case my preference would be to disable the generation of array allocators for classes for which an annotation has been placed on the info of the constructor (instead of just ignoring the info like it's the case currently). No need for additional Info.

saudet: Sure, until we can think of something useful to do, we can skip allocateArray in that case.

HGuillemet:
This PR was merged into #700.

PR #731, opened 2023-12-26
Upgraded versions for various plugins.

archenroot:
I upgraded few plugins and mostly:
```
 <artifactId>maven-compiler-plugin</artifactId>
149         <version>3.12.1</version>
150         <configuration>
151           <source>1.8</source>
152           <target>1.8</target>
```
as it was failing on my system to build with 1.7 values:
```
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=22.04
DISTRIB_CODENAME=jammy
DISTRIB_DESCRIPTION="Ubuntu 22.04.3 LTS"
java 21.0.1 2023-10-17
Java(TM) SE Runtime Environment Oracle GraalVM 21.0.1+12.1 (build 21.0.1+12-jvmci-23.1-b19)
Java HotSpot(TM) 64-Bit Server VM Oracle GraalVM 21.0.1+12.1 (build 21.0.1+12-jvmci-23.1-b19, mixed mode, sharing)

```

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -57,8 +57,8 @@
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <maven.build.timestamp.format>yyyyMMddhhmm</maven.build.timestamp.format>
-    <maven.version>3.0</maven.version>
-    <bnd.version>5.3.0</bnd.version>
+    <maven.version>3.9.0</maven.version>
+    <bnd.version>6.4.0</bnd.version>
     <javacpp.platform.root></javacpp.platform.root>
     <javacpp.platform.suffix></javacpp.platform.suffix>
     <javacpp.platform.compiler></javacpp.platform.compiler>
@@ -93,7 +93,7 @@
     <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-simple</artifactId>
-      <version>1.7.36</version>
+      <version>2.0.9</version>
       <optional>true</optional>
     </dependency>
     <!--
@@ -103,7 +103,7 @@
     <dependency>
       <groupId>org.osgi</groupId>
       <artifactId>osgi.annotation</artifactId>
-      <version>7.0.0</version>
+      <version>8.1.0</version>
       <scope>provided</scope>
     </dependency>
   </dependencies>
@@ -128,7 +128,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-plugin-plugin</artifactId>
-        <version>3.6.2</version>
+        <version>3.9.0</version>
         <configuration>
           <packagingTypes>
             <packagingType>jar</packagingType>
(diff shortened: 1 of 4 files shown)


saudet:
Could you do the same for https://github.com/bytedeco/javacv/ and https://github.com/bytedeco/javacpp-presets?

archenroot:
I was looking into javacv repo, but build fails on https://github.com/bytedeco/javacv/issues/2157

I will look into javacpp-presets as well.

archenroot:
I did the version bump for javacv: https://github.com/bytedeco/javacv/pull/2159. Going to look at javacpp-presets.

archenroot:
I reverted back the 1.8 to 1.7

pom.xml
```diff
@@ -57,8 +57,8 @@
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <maven.build.timestamp.format>yyyyMMddhhmm</maven.build.timestamp.format>
-    <maven.version>3.0</maven.version>
-    <bnd.version>5.3.0</bnd.version>
+    <maven.version>3.9.0</maven.version>
```
saudet, line 60: Upgrading this breaks compatibility with Maven 3.0.x, which isn't such big problem, but do you really need it?

saudet:
Please also update the dependencies in src/it/osgi/pom.xml, src/it/osgi/test.bndrun, platform/pom.xml and https://github.com/bytedeco/javacv/blob/master/platform/pom.xml and https://github.com/bytedeco/javacpp-presets/blob/master/platform/pom.xml as well

pom.xml
archenroot: not really. Do you mean just the maven.version or as well the bnd.version upgrade?

saudet: The maven.version, but I'm thinking we should upgrade that anyway, and revert at some point if it ever breaks something someone needs...

archenroot: ok, agreed.

saudet (approved):

codeinred:
@archenroot could you provide more information about the issue with groovy compatibility? by this point Java 1.7 is so old that Maven fails to build javacpp:

```
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR :
[INFO] -------------------------------------------------------------
[ERROR] Source option 7 is no longer supported. Use 8 or later.
[ERROR] Target option 7 is no longer supported. Use 8 or later.
[INFO] 2 errors
[INFO] -------------------------------------------------------------
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
```

I had to manually patch the repo in order to get it to build. Would it be possible to add a flag such that 1.8 was the default version, but 1.7 (or an alternative version) could be used for compatibility?

saudet:
We could make it 1.8 when we detect a new version of the JDK that doesn't support 1.7. Please open a pull request!

HGuillemet:
> @archenroot could you provide more information about the issue with groovy compatibility? by this point Java 1.7 is so old that Maven fails to build javacpp

The issue is with JavaCPP Gradle plugin only. See https://github.com/bytedeco/javacpp-presets/pull/1453#issuecomment-1907061433.

Issue #740, opened 2024-01-28
Multiple pointer types and automatic generation of function overloads 


HGuillemet:
If we have in C++:
```c++
void f(T t);
```
and:
```Java
new Info("T").pointerTypes("T1", "T2");
```
we generate:
```Java
void f(T1 t);
void f(T2 t);
```
If the function takes 2 arguments:
```c++
void f(T t, U u);
```
with:
```Java
new Info("T").pointerTypes("T1", "T2");
new Info("U").pointerTypes("U1", "U2");
```
we now generate:
```Java
void f(T1 t, U1 u);
void f(T2 t, U2 u);
```
Is this intentional ? 
Shouldn't we generate all combinations ? 
```Java
void f(T1 t, U1 u);
void f(T2 t, U1 u);
void f(T1 t, U2 u);
void f(T2 t, U2 u);
```

For instance in Pytorch, we have function `einsum` that takes as `std::string` and a `ArrayRef`.
We used to have 2 java methods, one taking `BytePointer` and `ArrayRef` and another `String` and `ArrayRef`.
But since I added `Vector` to pointer types for `c10::ArrayRef`, we now have a java method taking `BytePointer` and `ArrayRef` and another taking `String` and `Vector`. The overload taking `String` and `ArrayRef` is lost, which could break some user code.
```

**Comment by saudet** [comment]
If you want to generate more combinations, do something like this:
```java
new Info("T").pointerTypes("T1", "T2", "T1", "T2");
new Info("U").pointerTypes("U1", "U2", "U2", "U1");
```

PR #741, opened 2024-01-29
Check for name duplicates in containers

HGuillemet:
Sometimes we add duplicate names in `Info.pointerTypes` or `javaNames` to ensure that the proper method overloads are generated (see #740).

When a function is parsed, resulting Java declarations with same signature are deduplicated, but not when containers are generated.

This PR add checks to prevent duplicate generation of container constructors and methods in these cases.


HGuillemet:
Good to merge.

PR #748, opened 2024-03-15
fix: std::char_traits is deprecated for other type than char

Clement-Devevey:
Issue was the following:

`jniNativeLibrary.cpp:314:57: warning: 'char_traits<unsigned short>' is deprecated: char_traits<T> for T not equal to char, wchar_t, char8_t, char16_t or char32_t is non-standard and is provided for a temporary period. It will be removed in LLVM 18, so please migrate off of it. [-Wdeprecated-declarations]
    return JavaCPP_createStringFromUTF16(env, ptr, std::char_traits<unsigned short>::length(ptr));
                                                        ^
char_traits.h:79:8: note: 'char_traits<unsigned short>' has been explicitly marked deprecated here
struct _LIBCPP_DEPRECATED_("char_traits<T> for T not equal to char, wchar_t, char8_t, char16_t or char32_t is non-standard and is provided for a temporary period. It will be removed in LLVM 18, so please migrate off of it.")
       ^
c++/v1/__config:772:53: note: expanded from macro '_LIBCPP_DEPRECATED_'
#      define _LIBCPP_DEPRECATED_(m) __attribute__((deprecated(m)))`

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/tools/Generator.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Generator.java
@@ -1041,7 +1041,7 @@ public class Generator {
             out.println("    if (ptr == NULL) {");
             out.println("        return NULL;");
             out.println("    }");
-            out.println("    return JavaCPP_createStringFromUTF16(env, ptr, std::char_traits<unsigned short>::length(ptr));");
+            out.println("    return JavaCPP_createStringFromUTF16(env, ptr, strlen(reinterpret_cast<const char*>(ptr)));");
             out.println("}");
             out.println();
         }


Clement-Devevey:
Or     return JavaCPP_createStringFromUTF16(env, ptr, std::char_traits<char16_t>::length(reinterpret_cast<const char16_t*>(ptr)));
not sure what's best

saudet:
That doesn't sound like a correct fix. We need to know the length of the string in UTF-16.

char16_t is only available since C++11, so we need to use something else.

saudet:
Maybe we could just put an ifdef else endif in there that uses char16_t for C++11 compilers and keep using unsigned short if not?

saudet:
@Clement-Devevey Since we've decided to make C++11 a requirement for future versions of JavaCPP, let's just replace unsigned short with char16_t. Thanks!

Clement-Devevey:
> @Clement-Devevey Since we've decided to make C++11 a requirement for future versions of JavaCPP, let's just replace unsigned short with char16_t. Thanks!

what an exciting news, thanks a lot :)

saudet:
You'll open another PR with the change?

saudet:
Duplicate of #753

Issue #710, opened 2023-09-27
Warning with "JavaCPP_createStringFromUTF16" method on Mac clang 15


argonium:
I'm running clang / g++ v.15 on a MacBook.  When I compile my code with Java 11 and JavaCPP 1.5.9, I get this warning:

```
.../build/javacpp/jniNativeLexiconProviderLib.cpp:314:57: error: 'char_traits<unsigned short>' is deprecated: char_traits<T> for T not equal to char, wchar_t, char8_t, char16_t or char32_t is non-standard and is provided for a temporary period. It will be removed in LLVM 18, so please migrate off of it. [-Werror,-Wdeprecated-declarations]
    return JavaCPP_createStringFromUTF16(env, ptr, std::char_traits<unsigned short>::length(ptr));
                                                        ^
/Library/Developer/CommandLineTools/SDKs/MacOSX14.0.sdk/usr/include/c++/v1/__string/char_traits.h:79:8: note: 'char_traits<unsigned short>' has been explicitly marked deprecated here
struct _LIBCPP_DEPRECATED_("char_traits<T> for T not equal to char, wchar_t, char8_t, char16_t or char32_t is non-standard and is provided for a temporary period. It will be removed in LLVM 18, so please migrate off of it.")
       ^
/Library/Developer/CommandLineTools/SDKs/MacOSX14.0.sdk/usr/include/c++/v1/__config:836:53: note: expanded from macro '_LIBCPP_DEPRECATED_'
#      define _LIBCPP_DEPRECATED_(m) __attribute__((deprecated(m)))
                                                    ^
```

The warning seems to originate with this line of code from JavaCPP: https://github.com/bytedeco/javacpp/blob/master/src/main/java/org/bytedeco/javacpp/tools/Generator.java#L1043

The exact version of llvm on my Mac is:

```
$ llvm-g++ --version
Apple clang version 15.0.0 (clang-1500.0.40.1)
Target: x86_64-apple-darwin23.0.0
Thread model: posix
...
```

blueberry:
This is no longer a warning on new MacOS 15.4, but a compiler error so I guess JavaCPP 1.5.12. release is highly desirable when (if) this PR is accepted.

PR #753, opened 2024-04-05
Fix use of deprecated std::char_traits<unsigned short>::length

codeinred:
Some versions of clang - particularly the one on MacOS - emitted the following warning when compiling javacpp:

> `jniNativeLibrary.cpp:314:57: warning: 'char_traits' is deprecated:
> char_traits for T not equal to char, wchar_t, char8_t, char16_t or
> char32_t is non-standard and is provided for a temporary period. It
> will be removed in LLVM 18, so please migrate off of it.
> [-Wdeprecated-declarations] return JavaCPP_createStringFromUTF16(env,
> ptr, std::char_traits::length(ptr));

This commit fixes that warning by adding a function JavaCPP_stringLength for strings of unsigned short.

## Related Pull Requests & Issues

#748 was added with the same goal of fixing issue, however the concern raised there was that `char16_t` is not available pre-c++11. This PR aims to fix that issue by instead providing a clean implementation in standard C++.

If merged, this PR should resolve Issue #710. 

## Performance Implications

There should be no performance difference for this change versus using `std::chair_traits<unsigned short>::length`, because it appears that compilers do not handle `std::char_traits<unsigned short>::length` as a special case in the way they do `strlen`.

See [this godbolt link](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:31,endLineNumber:13,positionColumn:31,positionLineNumber:13,selectionStartColumn:31,selectionStartLineNumber:13,startColumn:31,startLineNumber:13),source:'%23include+%3Cmemory%3E%0A%23include+%3Cstring%3E%0A%0A%0Asize_t+char_traits_length(unsigned+short+const*+s)%0A%7B%0A++++return+std::char_traits%3Cunsigned+short%3E::length(s)%3B%0A%7D%0A%0A%0Asize_t+JavaCPP_stringLength(unsigned+short+const*+s)%0A%7B%0A++++for(size_t+i+%3D+0%3B+%3B+%2B%2Bi)+%7B%0A++++++++if+(s%5Bi%5D+%3D%3D+0)+return+i%3B%0A++++%7D%0A%7D%0A'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:g132,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:2,lang:c%2B%2B,libs:!(),options:'-O3+-Wall+-std%3Dc%2B%2B17',overrides:!(),selection:(endColumn:12,endLineNumber:21,positionColumn:1,positionLineNumber:1,selectionStartColumn:12,selectionStartLineNumber:21,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+gcc+13.2+(Editor+%231)',t:'0'),(h:output,i:(compilerName:'x64+msvc+v19.latest',editorid:1,fontScale:14,fontUsePx:'0',j:2,wrap:'0'),l:'5',n:'0',o:'Output+of+x86-64+gcc+13.2+(Compiler+%232)',t:'0')),header:(),k:50,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4) for reference.

Assembly output is identical between the standard library implementation, and the one provided in this commit.

**Implementation:**

```cpp
#include <memory>
#include <string>

/// Old implementation
size_t char_traits_length(unsigned short const* s)
{
    return std::char_traits<unsigned short>::length(s);
}

/// New implementation
size_t JavaCPP_stringLength(unsigned short const* s)
{
    for(size_t i = 0; ; ++i) {
        if (s[i] == 0) return i;
    }
}
```

**Assembly:**
```asm
char_traits_length(unsigned short const*):
        xor     eax, eax
        cmp     WORD PTR [rdi], 0
        je      .L4
.L3:
        add     rax, 1
        cmp     WORD PTR [rdi+rax*2], 0
        jne     .L3
        ret
.L4:
        ret
JavaCPP_stringLength(unsigned short const*):
        xor     eax, eax
        cmp     WORD PTR [rdi], 0
        je      .L7
.L8:
        add     rax, 1
        cmp     WORD PTR [rdi+rax*2], 0
        jne     .L8
.L7:
        ret
```

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/tools/Generator.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Generator.java
@@ -1001,6 +1001,17 @@ public class Generator {
         out.println("        env->SetLongField(obj, JavaCPP_capacityFID, (jlong)size);");
         out.println("    }");
         out.println("}");
+
+        // This is needed because char_traits<unsigned_short>::length is deprecated:
+        // warning: 'char_traits' is deprecated: char_traits for T not equal to char,
+        // wchar_t, char8_t, char16_t or char32_t is non-standard and is provided for
+        // a temporary period. It will be removed in LLVM 18, so please migrate off of it.
+        out.println("static JavaCPP_noinline size_t JavaCPP_stringLength(const unsigned short* ptr) {");
+        out.println("    for(size_t i = 0; ; ++i) {");
+        out.println("        if (ptr[i] == 0) return i;");
+        out.println("    }");
+        out.println("}");
+
         out.println();
         if (handleExceptions || convertStrings) {
             out.println("#include <string>");
@@ -1041,7 +1052,7 @@ public class Generator {
             out.println("    if (ptr == NULL) {");
             out.println("        return NULL;");
             out.println("    }");
-            out.println("    return JavaCPP_createStringFromUTF16(env, ptr, std::char_traits<unsigned short>::length(ptr));");
+            out.println("    return JavaCPP_createStringFromUTF16(env, ptr, JavaCPP_stringLength(ptr));");
             out.println("}");
             out.println();
         }
@@ -2568,7 +2579,7 @@ public class Generator {
                 // special considerations for char arrays as strings
                 if (asUtf16(methodInfo, k)) {
                     out.print(indent + "memcpy(");
-                    suffix = ", (std::char_traits<unsigned short>::length(ptr" + k + ") + 1) * sizeof(unsigned short))";
+                    suffix = ", (JavaCPP_stringLength(ptr" + k + ") + 1) * sizeof(unsigned short))";
                 } else {
                     out.print(indent + "strcpy((char*)");
                 }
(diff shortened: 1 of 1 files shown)


saudet:
Maybe we could just put an ifdef else endif in there that uses char16_t for C++11 compilers and keeps using unsigned short if not?

HGuillemet:
Either using a custom null-seeking function or an `ifdef` calling either `std::chair_traits<unsigned short>::length` or `std::chair_traits<char16_t>::length seem ok for me, since @codeinred established that there is no performance loss and the generated assembly is the same.

BTW, I see in usage of `wcslen` in `StringAdapter` when passed a `short *` or `int *`. That looks wrong.

saudet:
> BTW, I see usage of wcslen in StringAdapter when passed a short * or int *. That looks wrong.

That's just for platform dependent support of wchar, not UTF-16 or UTF-32

saudet:
@codeinred Since we've decided to make C++11 a requirement for future versions of JavaCPP, let's just replace unsigned short with char16_t. Thanks!

hehua2008:
Hope to merge this PR asap~

saudet:
@hehua2008 As soon as someone fixes the build!

saudet:
Superseded by #791

PR #755, opened 2024-05-18
Add libomp to windows preload

HGuillemet:
Add `libomp140` to the list of windows preload. This is the LLVM openMP runtime, which brings compatibility with OpenMP 3 and should become the standard for VS in replacement of `vsomp`.

Needed by Pytorch since 2.2.2.

Fixes bytedeco/javacpp-presets#1500.


saudet:
Yeah, that's not going to work. Please add both of them to the existing preload list.

saudet (approved):

PR #781, opened 2024-10-10
Support for riscv64

ixgbe00:
Support for riscv64 on LinuxOS

### DIFF
--- a/platform/pom.xml
+++ b/platform/pom.xml
@@ -66,8 +66,8 @@
     <javacpp.platform.ios-x86_64>ios-x86_64${javacpp.platform.extension}</javacpp.platform.ios-x86_64>
     <javacpp.platform.linux-armhf>linux-armhf${javacpp.platform.extension}</javacpp.platform.linux-armhf>
     <javacpp.platform.linux-arm64>linux-arm64${javacpp.platform.extension}</javacpp.platform.linux-arm64>
-    <javacpp.platform.linux-riscv64>linux-riscv64${javacpp.platform.extension}</javacpp.platform.linux-riscv64>
     <javacpp.platform.linux-ppc64le>linux-ppc64le${javacpp.platform.extension}</javacpp.platform.linux-ppc64le>
+    <javacpp.platform.linux-riscv64>linux-riscv64${javacpp.platform.extension}</javacpp.platform.linux-riscv64>
     <javacpp.platform.linux-x86>linux-x86${javacpp.platform.extension}</javacpp.platform.linux-x86>
     <javacpp.platform.linux-x86_64>linux-x86_64${javacpp.platform.extension}</javacpp.platform.linux-x86_64>
     <javacpp.platform.macosx-arm64>macosx-arm64${javacpp.platform.extension}</javacpp.platform.macosx-arm64>
@@ -134,13 +134,13 @@
       <groupId>org.bytedeco</groupId>
       <artifactId>javacpp</artifactId>
       <version>${project.version}</version>
-      <classifier>${javacpp.platform.linux-riscv64}</classifier>
+      <classifier>${javacpp.platform.linux-ppc64le}</classifier>
     </dependency>
     <dependency>
       <groupId>org.bytedeco</groupId>
       <artifactId>javacpp</artifactId>
       <version>${project.version}</version>
-      <classifier>${javacpp.platform.linux-ppc64le}</classifier>
+      <classifier>${javacpp.platform.linux-riscv64}</classifier>
     </dependency>
 <!--    <dependency>-->
 <!--      <groupId>org.bytedeco</groupId>-->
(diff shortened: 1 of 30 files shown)


saudet:
Please also update the workflow and update the GitHut Actions scripts to support that platform

saudet:
You'll need to update this first:
https://github.com/bytedeco/javacpp-presets/tree/master/.github/actions/deploy-ubuntu

ixgbe00:
> You'll need to update this first: https://github.com/bytedeco/javacpp-presets/tree/master/.github/actions/deploy-ubuntu

https://github.com/bytedeco/javacpp-presets/pull/1544

saudet:
You'll also need to add properties here so that the right compiler is used:
https://github.com/bytedeco/javacpp/tree/master/src/main/resources/org/bytedeco/javacpp/properties

src/main/resources/org/bytedeco/javacpp/properties/linux-riscv64.properties
```diff
@@ -0,0 +1,31 @@
+platform=linux-riscv64
+platform.path.separator=:
+platform.source.suffix=.cpp
+platform.includepath.prefix=-I
+platform.includepath=
+platform.compiler=g++
```
saudet, line 6: This needs to be to the compiler for linux-riscv64

ixgbe00: I added -march=rv64gc -mabi=lp64d for platform.compiler.default

saudet: I don't think GCC supports multiple targets like that. You'll need to give it the correct binary name.

ixgbe00: platform.compiler should be riscv64-linux-gnu-g++  ,  however where should I  install riscv64 cross-compilation toolchain by the command  ' sudo apt install g++-riscv64-linux-gnu '  ,  could you tell me ? Thanks to you!

saudet: It should already be installed, just give it a try

ixgbe00: It seems that https://github.com/bytedeco/javacpp-presets/pull/1544 does not work
image
image

saudet: Ok, I fixed that, but now we're getting this error:
```
Some packages could not be installed. This may mean that you have
requested an impossible situation or if you are using the unstable
distribution that some required packages have not yet been created
or been moved out of Incoming.
The following information may help to resolve the situation:
The following packages have unmet dependencies:
 openjdk-8-jdk : Depends: openjdk-8-jre (= 8u422-b05-1~20.04) but it is not going to be installed
E: Unable to correct problems, you have held broken packages.
```
That needs to be fixed

saudet (changes requested):
Please make the build pass

ixgbe00:
> Please make the build pass
I've researched it for a while, but haven't found any good solutions. Could you give me some advice?

saudet:
Looks like some issue with Ubuntu's repository for the JDK and riscv64...

ixgbe00:
> Looks like some issue with Ubuntu's repository for the JDK and riscv64...


It seems that  java-8-openjdk-amd64 is used on ppc64el

ixgbe00:
> Looks like some issue with Ubuntu's repository for the JDK and riscv64...

sudo apt install openjdk-11-jdk -y
I can install  openjdk-11-jdk on riscv64 platform , but can not install  openjdk-8-jdk. Therefore if we can use openidk11 ? 
image

saudet:
I'm not going to accept that until next year, but you can maintain your own fork until then I guess

ixgbe00:
> I'm not going to accept that until next year, but you can maintain your own fork until then I guess

I get it.Thanks for your advice.

src/main/resources/org/bytedeco/javacpp/properties/linux-riscv64.properties
saudet: Ok, I've updated the builds to Jammy and that seems to work now. So this looks good to merge, but could you please fix the formatting first? Thank you!

platform/pom.xml
```diff
@@ -66,6 +66,7 @@
     <javacpp.platform.ios-x86_64>ios-x86_64${javacpp.platform.extension}</javacpp.platform.ios-x86_64>
     <javacpp.platform.linux-armhf>linux-armhf${javacpp.platform.extension}</javacpp.platform.linux-armhf>
     <javacpp.platform.linux-arm64>linux-arm64${javacpp.platform.extension}</javacpp.platform.linux-arm64>
+	<javacpp.platform.linux-riscv64>linux-riscv64${javacpp.platform.extension}</javacpp.platform.linux-riscv64>
```
saudet, line 69: Please fix formatting like here and all places below!

saudet (approved):

PR #727, opened 2023-11-18
Updated Generator.java to fix the flaky errors

njain2208:
**PR Overview:**
_________________________________________________________________________________________________________
This PR fixes the flaky/non-deterministic behavior of the following test because it assumes the ordering of methods returned by `getDeclaredMethods()`

[org.bytedeco.javacpp.BufferTest](https://github.com/bytedeco/javacpp/blob/master/src/test/java/org/bytedeco/javacpp/BufferTest.java)

[org.bytedeco.javacpp.EnumTest](https://github.com/bytedeco/javacpp/blob/master/src/test/java/org/bytedeco/javacpp/EnumTest.java)

[org.bytedeco.javacpp.PointerTest](https://github.com/bytedeco/javacpp/blob/master/src/test/java/org/bytedeco/javacpp/PointerTest.java)

**Test Overview:**
_________________________________________________________________________________________________________
In all of the above tests, the tests call the `setUpClass()` method where the problem arises. The `setUpClass()` makes an internal to the `Generator` class’s object where the problem resides.

You can reproduce the issue by running the following commands (the below command is for BufferTest, please change class name for targeting other tests) -

```
mvn install -pl . -am -DskipTests
mvn test  -Dtest=org.bytedeco.javacpp.BufferTest
mvn edu.illinois:nondex-maven-plugin:2.1.1:nondex -Dtest=org.bytedeco.javacpp.BufferTest
```

**Root Cause**
_________________________________________________________________________________________________________

`Method[] methods = cls.getDeclaredMethods();`

[...]

PR #784, opened 2024-10-24
Refined Generator.java to fix a flaky error

Ethan-Chin:
## Description
This PR fixes a flaky error in `setUpClass()` caused by inconsistent ordering of methods returned by two separate `getDeclaredMethods()` calls, as described in https://github.com/bytedeco/javacpp/pull/727.

The previous solution in https://github.com/bytedeco/javacpp/pull/727 sorted methods by name, which added inefficiency. Instead, this PR merges the two `getDeclaredMethods()` calls into one, ensuring a consistent method order without the need for sorting. (https://github.com/bytedeco/javacpp/pull/727#issuecomment-1826464058)

## Key Changes
Modified the function `static Method[] functionMethods` to accept the methods array as a parameter, avoiding the redundant call to `getDeclaredMethods()` within the function that lead to inconsistency.

## Related Issues
https://github.com/bytedeco/javacpp/pull/727

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/tools/Generator.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Generator.java
@@ -2116,7 +2116,7 @@ public class Generator {
             c = c.getSuperclass();
         }
         boolean[] callbackAllocators = new boolean[methods.length];
-        Method[] functionMethods = functionMethods(cls, callbackAllocators);
+        Method[] functionMethods = functionMethods(cls, methods, callbackAllocators);
         boolean firstCallback = true;
         for (int i = 0; i < methods.length; i++) {
             if (!Loader.checkPlatform(methods[i].getAnnotation(Platform.class), properties)) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/bytedeco/javacpp/tools/Generator.java
```diff
@@ -4386,7 +4385,7 @@ String[] cppTypeName(Class<?> type, Annotation[] annotations) {
         } else if (type.isPrimitive()) {
             prefix = type.getName();
         } else if (FunctionPointer.class.isAssignableFrom(type)) {
-            Method[] functionMethods = functionMethods(type, null);
+            Method[] functionMethods = functionMethods(type, type.getDeclaredMethods(), null);
```
saudet, line 4388: Doesn't that just move the call to getDeclaredMethods() outside of functionMethods()? Why does this fix anything?

saudet: Ah, it's not for this call, but the one above, got it.

saudet (approved):

saudet:
Looks good, thanks for the fix!

PR #791, opened 2025-01-20
fix: deprecated use of std::char_traits<unsigned short>::length

Clement-Devevey:
Fix the warning about deprecated use of char_traits<unsigned int> by  char_traits<char16_t> as it has been discussing in this thread: 
@codeinred Since we've decided to make C++11 a requirement for future versions of JavaCPP, let's just replace unsigned short with char16_t. Thanks!

_Originally posted by @saudet in https://github.com/bytedeco/javacpp/issues/753#issuecomment-2267543152_

--
Original issue is;
warning: 'char_traits<unsigned short>' is deprecated: char_traits<T> for T not equal to char, wchar_t, char8_t, char16_t or char32_t is non-standard and is provided for a temporary period. It will be removed in LLVM 19, so please migrate off of it. [-Wdeprecated-declarations]

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/tools/Generator.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Generator.java
@@ -1042,7 +1042,7 @@ public class Generator {
             out.println("    if (ptr == NULL) {");
             out.println("        return NULL;");
             out.println("    }");
-            out.println("    return JavaCPP_createStringFromUTF16(env, ptr, std::char_traits<unsigned short>::length(ptr));");
+            out.println("    return JavaCPP_createStringFromUTF16(env, ptr, std::char_traits<char16_t>::length(ptr));");
             out.println("}");
             out.println();
         }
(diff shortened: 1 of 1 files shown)


Clement-Devevey:
Having one issue about doing this: 
in `JavaCPP_createStringFromUTF16` the  `return env->NewString(ptr, length);` isn't working:
Cannot initialize a parameter of type 'const jchar *' (aka 'const unsigned short *') with an lvalue of type 'const char16_t *'

Clement-Devevey:
Apparently this is safe to cast from char16_t to jchar

It is generally safe to cast char16_t* to jchar* in JNI, as both types are 16-bit wide and represent UTF-16 encoded characters. This approach is commonly used in JNI code to handle UTF-16 strings.

Here’s a brief explanation:

char16_t: A 16-bit character type introduced in C++11 to represent UTF-16 encoded characters.
jchar: A 16-bit type used by JNI to represent UTF-16 encoded characters, which is typically defined as unsigned short.
Since both char16_t and jchar are 16-bit wide and represent UTF-16 characters, casting between them is safe and does not alter the data. This practice is used in various JNI implementations.µ

--

Using reinterpret_cast ensures that the conversion is done correctly at the bit level, which is necessary for this specific case in JNI where char16_t and jchar are essentially the same type but with different names.

saudet:
There a few other places where we could switch from `unsigned short` to `char16_t`, so we might as well do it now?

saudet:
It looks like it doesn't work at all if we don't change all `unsigned short` to `char16_t`, so please do that! Thanks

saudet:
Please fix the builds

saudet:
Superseded by https://github.com/bytedeco/javacpp/pull/800

PR #798, opened 2025-03-18
Advanced C++ -> Java mapping

Avdiji:
Added serveral functionalities which enable the user to map more advanced and specific edgecases. 
Here's a list of all the additional features:

1. Custom Exception-Mapping
   - Using the ExceptionMapper annotation within the Platform annotation you can now map custom C++ exceptions to any java Throwable.
   -   Just specify the CPP exception (as a String) and the corresponding java exception and the mapper will automatically catch any CPP exception of that type, unroll it recursively and map it to the given java exception.

2. Custom JNI-Code in the beggining of each generated JNI-File
    - Within the Platform annotation there now is a new property 'addCustomJNIFrom' which allows the user to specify the resource-path to any file. The files content will then be inserted on top of the generated JNI-Code.

3. Custom JNI-Mapping of individual java methods
    - The CustomMapper annotation allows the user to write his own JNI code for any function. Attirbutes are:
        - customMapping: A String which will directly replace any otherwhise generated code within the try-catch block of the mapping function
        - filePaths: A String[] which reflects the resourcePaths of the files which contain the mapping. The files content will replace the otherwhise generated code within the try-cathc block of the mapping function, in the order of the values within the String[]
        - functionCall: will replace the String '$funcName'
        - passCTypeParams: determines whether to pass the generated field ptr or the parameter arg into the calling function.
        - dereferenceParams: will dereference any occurence of 'ptr' wihtin the calling function.
        - replacements: a String[] with replacements (e.g.: ["val1", "val2", "val3", "val4"] -> val1 will be replaced by val2 and val3 by val4...)
        - typename: will change the typename of the underlying object to whatever passed String.
       
4. Added several properties to the Virtual annotation, in order to enable Cross-Language Polymorphism for custom mapped datatypes.
5. Added setClassLoader function, which ensures a ClassLoader at everytime (using it is optional though).

Overall the changes are supposed to allow the user to map C++ code in a more specific way and gives more freedom to add custom JNI-Code.


Avdiji:
Sure thing! I'll get to it once i find the time

saudet:
No need to do all of them at once, just a first small pull request that stands alone, and then when that gets merged, another small pull request based on the first set of changes, etc. Thanks!

PR #800, opened 2025-04-08
Fix #710.

### LINKED ISSUE #710: Warning with "JavaCPP_createStringFromUTF16" method on Mac clang 15

blueberry:
This PR fixes #710 .

With the release of MacOS 15.4, and the corresponding default C++ compiler upgrade, what was previously a warning, now is a compiler error, so this fixes a show stopper on new macos machine builds. On older systems, it removes the warnings, I suppose :) 

Build passes on all OSs defined in GitHub actions.


pom.xml
```diff
@@ -148,8 +148,8 @@
         <artifactId>maven-compiler-plugin</artifactId>
         <version>3.12.1</version>
         <configuration>
-          <source>1.7</source>
-          <target>1.7</target>
+          <source>1.8</source>
+          <target>1.8</target>
```
saudet, line 152: Let's leave that at 1.7 for now...

src/main/java/org/bytedeco/javacpp/tools/Generator.java
```diff
@@ -3028,7 +3028,7 @@ void parametersAfter(MethodInformation methodInfo) {
             if ("void*".equals(typeName[0]) && !methodInfo.parameterTypes[j].isAnnotationPresent(Opaque.class)) {
                 typeName[0] = "char*";
             }
-            
+
```
saudet, line 3031: Please revert unnecessary changes in formatting here and there

src/main/java/org/bytedeco/javacpp/tools/Generator.java
```diff
@@ -2327,7 +2327,7 @@ void parametersBefore(MethodInformation methodInfo) {
                     String s = methodInfo.parameterTypes[j].getComponentType().getName();
                     if (methodInfo.criticalRegion || methodInfo.valueGetter || methodInfo.valueSetter ||
                             methodInfo.memberGetter || methodInfo.memberSetter) {
-                        out.println("(j" + s + "*)env->GetPrimitiveArrayCritical(arg" + j + ", NULL);");
+                        out.println("reinterpret_cast<" + typeName[0] + ">(env->GetPrimitiveArrayCritical(arg" + j + ", NULL));");
```
saudet, line 2330: Why do we need to change this here?

src/main/java/org/bytedeco/javacpp/tools/Generator.java
saudet: Although this looks like it's removing white space, so that's good thing. No need to revert

pom.xml
blueberry: @saudet I wish I could, but maven on my machines refuses to build 1.7 source/target.

src/main/java/org/bytedeco/javacpp/tools/Generator.java
blueberry: Because the C++ compiler refuses to cast jchar * to char16_t * (or something along this line, I didn't save the exact error message at the moment). Anyway, its necessary for the unsigned short -> char16_t transision.

src/main/java/org/bytedeco/javacpp/tools/Generator.java
blueberry: Yes.

src/main/java/org/bytedeco/javacpp/tools/Generator.java
saudet: Well, let's see what else we can do, because this doesn't look correct...

saudet: In the lines below, `(" + typeName[0] + typeName[1] + ")` get used, so if that works let's do that

blueberry: You mean with or without reinterpret_cast?

saudet: Without

blueberry: OK. It works (on my macos).

saudet (approved):

karllessard:
Hey @saudet , any plan for releasing this fix soon? Thanks

saudet:
@karllessard That only affects compilation, not runtime

karllessard:
Yes but we were in the process of upgrading to TensorFlow 2.18 and this is blocking us right now, since I need to compile locally once for getting the ops generated.

saudet:
We can easily use the snapshots:
http://bytedeco.org/builds/

jkhoogland:
Hi Samuel,
it would be great if we could a new release for this out.
Thanks,
Jiri

saudet:
Yeah, well, NVIDIA seems to think otherwise, see https://github.com/bytedeco/javacpp-presets/issues/1620#issuecomment-2953669875

blueberry:
Is it technically possible to release 1.5.12 for everything except Pytorch and OpenCV, and then release these two (as 1.5.12, or 1.5.13) as soon as the issue is fixed?

saudet:
Sure, anything is possible, but it's not something I will personally take the time to deal with. It's not like NVIDIA doesn't have enough money to help us out now do they. 😮‍💨

blueberry:
Sure, but they don't care about us.  Maybe they'd even care about this issue if we could reach someone in the trenches and let them know that the issue exists, but I unfortunately don't have any contact...

blueberry:
As for the help to this project, I feel your frustration. But I do think that you're too hesitant of explicitly asking for financial support by companies. Yes, GitHub offers easy way to sponsor a project on a small scale, but companies especially don't care about something lke that unless someone pushes the people in seeing this as an issue. Maybe there should be an explicit plea for funding, with explicit reasoning, ways to contribute, budget needed, expected amounts, etc. I'm sure that wouldn't send tons of money towards your efforts with JavaCPP, but I'm pretty sure there would be *some* funds, and, who knows, maybe a decent support after some time. I understand that you may be frustrated enough with the cruel corporate world that you don't want to go that route, and that you've tried many times in the past with insufficient success, but times have changed in the past decade I believe. Anyway, currently I don't see any prominent call to action related to that on the JavaCPP website, so even people who could help, just don't think about that and don't even know that you need it.

saudet:
It's all politics at this point, and I don't do politics. If the Java community had a need of JavaCPP, there would be demand for that kind of tool out there, but there isn't, not anymore, Oracle made sure of that. Anyway, I've been able to fix the builds in a satisfactory fashion, so if there are no other major issues, we should be able to make a release by the end of the month.

PR #812, opened 2025-07-20
Fix ndk28 compile

tmm1:
cc https://github.com/bytedeco/javacpp/pull/244/files#r2217665136


saudet:
Those properties are deprecated. Please update android-arm.properties and android-x86.properties instead.

tmm1:
The other files don't have any references to android_support

saudet:
https://github.com/bytedeco/javacpp/blob/master/src/main/resources/org/bytedeco/javacpp/properties/android-arm.properties#L29
https://github.com/bytedeco/javacpp/blob/master/src/main/resources/org/bytedeco/javacpp/properties/android-x86.properties#L29

tmm1:
Sorry I was looking at main from 2020 🤦

saudet (approved):

PR #816, opened 2025-07-29
Introduce `NativeAllocationTracer` for tracking native memory allocation sites

devjeonghwan:
Related to #815 

## Example
```java
    @Test
    public void testTracer() {
        FloatPointer fp = new FloatPointer(128);

        System.out.println("=======================");
        for (NativeAllocationTracer.Site site : NativeAllocationTracer.getSites()) {
            System.out.println(site.toString());
        }

        fp = null;
        for (int i= 0; i < 128; i++) {
            System.gc();
            Thread.yield();
        }

        System.out.println("=======================");
        for (NativeAllocationTracer.Site site : NativeAllocationTracer.getSites()) {
            System.out.println(site.toString());
        }

        fp = new FloatPointer(128);

        System.out.println("=======================");
        for (NativeAllocationTracer.Site site : NativeAllocationTracer.getSites()) {
            System.out.println(site.toString());
        }

        fp.deallocate();

        System.out.println("=======================");
        for (NativeAllocationTracer.Site site : NativeAllocationTracer.getSites()) {
            System.out.println(site.toString());
        }
    }
```

## Output
```
=======================
Site[location=org.bytedeco.javacpp.TracerTest.testTracer(TracerTest.java:18),totalCounts=1,totalBytes=512,liveCounts=1,liveBytes=512,collectedCounts=0,collectedBytes=0]
=======================
Site[location=org.bytedeco.javacpp.TracerTest.testTracer(TracerTest.java:18),totalCounts=1,totalBytes=512,liveCounts=0,liveBytes=0,collectedCounts=1,collectedBytes=512]
=======================
Site[location=org.bytedeco.javacpp.TracerTest.testTracer(TracerTest.java:36),totalCounts=1,totalBytes=512,liveCounts=1,liveBytes=512,collectedCounts=0,collectedBytes=0]
Site[location=org.bytedeco.javacpp.TracerTest.testTracer(TracerTest.java:18),totalCounts=1,totalBytes=512,liveCounts=0,liveBytes=0,collectedCounts=1,collectedBytes=512]
=======================
Site[location=org.bytedeco.javacpp.TracerTest.testTracer(TracerTest.java:36),totalCounts=1,totalBytes=512,liveCounts=0,liveBytes=0,collectedCounts=0,collectedBytes=0]
Site[location=org.bytedeco.javacpp.TracerTest.testTracer(TracerTest.java:18),totalCounts=1,totalBytes=512,liveCounts=0,liveBytes=0,collectedCounts=1,collectedBytes=512]
```

### DIFF
--- a/src/main/java/org/bytedeco/javacpp/Pointer.java
+++ b/src/main/java/org/bytedeco/javacpp/Pointer.java
@@ -128,7 +128,7 @@ public class Pointer implements AutoCloseable {
                 try {
                     tracerMarkPointerMethod.invoke(null, this);
                 } catch (ReflectiveOperationException exception) {
-                    logger.error("Unable to invoke native allocation tracer method via reflection: " + exception);
+                    logger.error("Unable to invoke NativeAllocationTracer method via reflection: " + exception);
                 }
             }
             deallocator(new NativeDeallocator(this, ownerAddress, deallocatorAddress));
@@ -231,7 +231,7 @@ public class Pointer implements AutoCloseable {
                     tracerMarkReferenceMethod.invoke(null, this, p);
                     tracerRecordAllocationMethod.invoke(null, this, this.bytes);
                 } catch (ReflectiveOperationException exception) {
-                    logger.error("Unable to invoke native allocation tracer method via reflection: " + exception);
+                    logger.error("Unable to invoke NativeAllocationTracer method via reflection: " + exception);
                 }
             }
         }
@@ -250,7 +250,7 @@ public class Pointer implements AutoCloseable {
                     try {
                         tracerRecordDeallocationMethod.invoke(null, this, this.bytes);
                     } catch (ReflectiveOperationException exception) {
-                        logger.error("Unable to invoke native allocation tracer method via reflection: " + exception);
+                        logger.error("Unable to invoke NativeAllocationTracer method via reflection: " + exception);
                     }
                 }
             }
@@ -418,7 +418,7 @@ public class Pointer implements AutoCloseable {
                             try {
                                 tracerRecordCollectionMethod.invoke(null, deallocator, r.bytes);
                             } catch (ReflectiveOperationException exception) {
-                                logger.error("Unable to invoke native allocation tracer method via reflection: " + exception);
+                                logger.error("Unable to invoke NativeAllocationTracer method via reflection: " + exception);
                             }
                         }
                     }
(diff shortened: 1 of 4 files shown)


devjeonghwan:
I like the idea too! I was actually considering moving it to the `tools` package as well. However, this creates accessibility issues for all the `mark`/`record` methods.

The problem with your suggestion is that we'd need to handle 5 methods. Even if we get all the Method objects via reflection during initial time, writing all that verbose reflection code at every call site seems weird.

Making those methods to `public` would solve the accessibility issue, but then we got pollution problem from external calls, which isn't an easy decision either...

I'm curious is there a compelling reason why it must be moved to `tools`? What specific benefits would we gain that outweigh these complications?

saudet:
Putting everything we don't need at runtime in a difference package simply avoids namespace pollution, that's all.

And if we don't link dynamically, then the JVM needs to be able to load that class even when it's not needed.

saudet:
I suppose we could leave it as package private, that also prevents namespace pollution, but we still need a way to load it dynamically. In the case when we don't need to add "import" statements, does the JVM still loads all classes that might not even get used?

saudet:
Huh, I guess not. With Foo.java and Bar.java like this

```
public class Foo {
        System.out.println("foo");
        if (args.length > 0)
            Bar.bar();
    }
}
```
```
class Bar {
    static void bar() { System.out.println("bar"); }
}
```
```
$ javac Foo.java Bar.java
$ java Foo bar
foo
bar
$ rm Bar.class
$ java Foo bar
foo
Exception in thread "main" java.lang.NoClassDefFoundError: Bar
	at Foo.main(Foo.java:6)
Caused by: java.lang.ClassNotFoundException: Bar
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:641)
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188)
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:526)
	... 1 more
$ java Foo
foo
```
So I guess having a package private class that doesn't need any imports is fine after all

saudet:
However, it's not possible to tell AOT tools like GraalVM to keep those debug classes out of the image. When we use Class.forName, that becomes possible. I guess that's the reason we should use dynamic loading explicitly

devjeonghwan:
What's the status on this PR? Anything else I should work on?

src/main/java/org/bytedeco/javacpp/Pointer.java
```diff
@@ -124,6 +125,13 @@ void init(long allocatedAddress, long allocatedCapacity, long ownerAddress, long
         limit = allocatedCapacity;
         capacity = allocatedCapacity;
         if (ownerAddress != 0 && deallocatorAddress != 0) {
+            if (nativeAllocationTracerEnabled) {
+                try {
+                    tracerMarkPointerMethod.invoke(null, this);
+                } catch (IllegalAccessException | InvocationTargetException exception) {
```
saudet, line 131: Let's just catch ReflectiveOperationException for simplicity

devjeonghwan: Oh, thanks. Applied.

PR #829, opened 2026-05-19
Prevent Parser from identifying certain operator function as templates (#828)

### LINKED ISSUE #828: Fails to parse some operator functions
When generating a bindings class, some function declaration can cause JavaCPP to fail to a completely valid header file. 

The problematic header is from CEF 147.0.14+g76d2442, and it is located in `include/internal/cef_string_wrappers.h`
Here is the snippet that caused a failure:
```h
//more on top...


// These functions are provided as a convenience for logging, which is where we
// use streams. Non-ASCII characters will be converted to UTF-8.
template <class traits>
inline std::ostream& operator<<(std::ostream& out,
                                const CefStringBase<traits>& str) {
  return operator<<(out, str.ToString());
}

#endif  // CEF_INCLUDE_INTERNAL_CEF_STRING_WRAPPERS_H_
```

And here is the log:
```log
Info: Targeting src\generator\java\com\tiji\silcef\Natives.java
Info: Parsing cef\include\internal\cef_string_wrappers.h
Exception in thread "main" org.bytedeco.javacpp.tools.ParserException: null:0:Unexpected token 'EOF'
	at org.bytedeco.javacpp.tools.Token.expect(Token.java:114)
	at org.bytedeco.javacpp.tools.Parser.templateArguments(Parser.java:779)
	at org.bytedeco.javacpp.tools.Parser.type(Parser.java:840)
	at org.bytedeco.javacpp.tools.Parser.type(Parser.java:809)
	at org.bytedeco.javacpp.tools.Parser.declarations(Parser.java:4476)
	at org.bytedeco.javacpp.tools.Parser.parse(Parser.java:4614)
	at org.bytedeco.javacpp.tools.Parser.parse(Parser.java:4764)
	at org.bytedeco.javacpp.tools.Builder.parse(Builder.java:95)
	at org.bytedeco.javacpp.tools.Builder.build(Builder.java:1095)
	at org.bytedeco.javacpp.tools.Builder.main(Builder.java:1450)
```

I think parser is confused by call to `operator<<`, but it is only my thoughts, and I know nothing about internals.

IamTiji:
This fixes #828. 

It adds an additional check before handling `<` token, correctly handling call to `operator<<`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 
+ * Fix `Parser` failing on `template<... operator<...` functions ([pull #829](https://github.com/bytedeco/javacpp/pull/829))
  * Compile classes with `parameters` bumping minimum requirements to Java SE 8 and Android 7.0 ([issue bytedeco/javacpp-presets#1739](https://github.com/bytedeco/javacpp-presets/issues/1739))
 
 ### February 22, 2026 version 1.5.13
--- a/src/main/java/org/bytedeco/javacpp/tools/Parser.java
+++ b/src/main/java/org/bytedeco/javacpp/tools/Parser.java
@@ -836,7 +836,8 @@ public class Parser {
                 }
                 tokens.next();
                 break;
-            } else if (token.match('<')) {
+            // If it is an operator, it is probably `operator<<`, not start of a template
+            } else if (token.match('<') && !type.operator) {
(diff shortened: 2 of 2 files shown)


saudet (approved):

saudet:
Thanks for the contribution!

